PR #46, opened 2024-05-01
Make ollama_client using Ollama OpenAI compatible API

guangrui-cloudnatix:
Ollama go sdk does not support chat using OpenAI compatible API, which is recently introduced in Ollama. Copy related functions from Ollama SDK and alternate it to process interfence-manager-server messages

### DIFF
--- a/server/internal/config/config.go
+++ b/server/internal/config/config.go
@@ -9,8 +9,10 @@ import (
 
 // Config is the configuration.
 type Config struct {
-	GRPCPort int `yaml:"grpcPort"`
-	HTTPPort int `yaml:"httpPort"`
+	GRPCPort         int    `yaml:"grpcPort"`
+	HTTPPort         int    `yaml:"httpPort"`
+	MonitoringPort   int    `yaml:"monitoringPort"`
+	OllamaServerAddr string `yaml:"ollamaServerAddr"`
 
 	AuthConfig AuthConfig `yaml:"auth"`
 }
@@ -39,6 +41,12 @@ func (c *Config) Validate() error {
 	}
 	if c.HTTPPort <= 0 {
 		return fmt.Errorf("httpPort must be greater than 0")
+	}
+	if c.MonitoringPort <= 0 {
+		return fmt.Errorf("monitoringPort must be greater than 0")
+	}
+	if c.OllamaServerAddr == "" {
+		return fmt.Errorf("ollamaServerAddr must be set")
 	}
 	if err := c.AuthConfig.Validate(); err != nil {
 		return err
--- a/server/internal/server/ollama_client.go
+++ b/server/internal/server/ollama_client.go
@@ -1,72 +1,68 @@
 package server
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
+	"io"
 	"log"
+	"net/http"
+	"net/url"
 
-	v1 "github.com/llm-operator/inference-manager/api/v1"
-	"github.com/ollama/ollama/api"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
 
-func newClient() (*api.Client, error) {
-	client, err := api.ClientFromEnvironment()
+type client struct {
+	base *url.URL
+	http *http.Client
+}
+
+func newClient(ollamaServerAddr string) *client {
+	return &client{
+		base: &url.URL{
+			Scheme: "http",
+			Host:   ollamaServerAddr,
+		},
+		http: http.DefaultClient,
+	}
+}
+
+func (c *client) sendRequest(ctx context.Context, method, path string, data any) ([]byte, error) {
+	var buf *bytes.Buffer
+	bts, err := json.Marshal(data)
 	if err != nil {
 		return nil, err
 	}
-	return client, nil
-}
+	buf = bytes.NewBuffer(bts)
 
-func handleChatRequest(ctx context.Context, req *v1.CreateChatCompletionRequest) (*v1.ChatCompletion, error) {
-	client, err := newClient()
+	requestURL := c.base.JoinPath(path)
+	request, err := http.NewRequestWithContext(ctx, method, requestURL.String(), buf)
 	if err != nil {
-		return nil, status.Errorf(codes.Internal, "Failed to create a client: %s", err)
+		return nil, err
 	}
 
-	var msgs []api.Message
-	for _, msg := range req.Messages {
-		msgs = append(msgs, api.Message{
-			Role:    msg.Role,
-			Content: msg.Content,
-		})
+	request.Header.Set("Content-Type", "application/json")
+	request.Header.Set("Accept", "application/x-ndjson")
+
+	response, err := c.http.Do(request)
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		if err := response.Body.Close(); err != nil {
+			log.Printf("Failed to close the response body: %s\n", err)
+		}
+	}()
+
+	if response.StatusCode < http.StatusOK || response.StatusCode >= http.StatusBadRequest {
+		log.Printf("Received an error response: %+v\n", response)
+		return nil, status.Errorf(codes.Code(response.StatusCode), "Status: %s", response.Status)
 	}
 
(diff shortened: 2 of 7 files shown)


server/internal/server/ollama_client.go
```diff
@@ -1,72 +1,133 @@
...
+type client struct {
+	base *url.URL
+	http *http.Client
+}
+
+func newClient() *client {
+	return &client{
+		base: &url.URL{
+			Scheme: "http",
+			Host:   os.Getenv("OLLAMA_HOST"),
```
kkaneda, line 32: Can't this be just a parameter to `newClient`?

server/internal/server/ollama_client.go
```diff
@@ -1,72 +1,133 @@
...
+	return nil
+}
+
+// chatResponseFunc is a function that [client.Chat] invokes every time
+// a response is received from the service. If this function returns an error,
+// [client.Chat] will stop generating and return this error.
+type chatResponseFunc func(*v1.ChatCompletion) error
+
+// Chat generates the next message in a chat.
+func (c *client) Chat(ctx context.Context, req *v1.CreateChatCompletionRequest, fn chatResponseFunc) error {
```
kkaneda, line 102:
```suggestion
func (c *client) chat(ctx context.Context, req *v1.CreateChatCompletionRequest, fn chatResponseFunc) error {
```

server/internal/server/ollama_client.go
```diff
@@ -1,72 +1,133 @@
...
+		}
+
+		return fn(&resp)
+	})
+}
+
+func handleChatRequest(ctx context.Context, req *v1.CreateChatCompletionRequest) (*v1.ChatCompletion, error) {
+	client := newClient()
+
+	var resp *v1.ChatCompletion
```
kkaneda, line 116: It seems to me that this works only when `fn` is called only once? I thought `fn` can be called multiple times.

server/internal/server/ollama_client.go
```diff
@@ -1,72 +1,133 @@
...
 }
 
-func handleChatRequest(ctx context.Context, req *v1.CreateChatCompletionRequest) (*v1.ChatCompletion, error) {
-	client, err := newClient()
+const maxBufferSize = 512 * 1000
+
+func (c *client) stream(ctx context.Context, method, path string, data any, fn func([]byte) error) error {
+	var buf *bytes.Buffer
+	if data != nil {
+		bts, err := json.Marshal(data)
```
kkaneda, line 43: I might not be reading the code right, but will this work? I thought we're passing a proto message, not json.

server/internal/server/ollama_client.go
guangrui-cloudnatix: not very sure what you mean by "called only once"? Here `processChatResponse` is expected to be a read-only function of `resp`, so no need to create a copy of original `resp`.

server/internal/server/ollama_client.go
guangrui-cloudnatix: seems working

```
kubectl port-forward -n llm-operator services/llm-operator-inference-manager-server-http 8444 &

curl http://localhost:8444/v1/chat/completions -d '{
  "model": "google-gemma-2b-it-q4",
  "messages": [{"role": "user", "content": "Why is the sky blue?"}]
}'
{"id":"chatcmpl-246", "choices":[{"finishReason":"stop", "index":0, "message":{"content":"The sky appears blue due to Rayleigh scattering. This scattering of light occurs when light interacts with molecules in the Earth's atmosphere.\n\n**Rayleigh scattering** is a process in which light waves are scattered in different directions by tiny particles in the medium through which they travel.\n\nIn the case of the sky, the molecules that scatter light are molecules of nitrogen and oxygen in the air. These molecules are much smaller than the wavelengths of visible light, which means that they scatter blue light more efficiently than other colors.\n\nWhen sunlight enters the atmosphere, it interacts with these molecules and is scattered in all directions. The blue light is scattered more strongly due to its shorter wavelength, resulting in the blue color that we see in the sky.\n\nOther factors that can influence the perceived color of the sky include:\n\n* **Dust particles:** Dust particles in the atmosphere can also scatter light, which can affect the color of the sky.\n* **Clouds:** Clouds can also scatter light, which can change the color of the sky.\n* **Sun position:** The position of the sun relative to the observer can also influence the color of the sky.\n\nOverall, the sky appears blue due to Rayleigh scattering of light by atmospheric molecules.", "toolCalls":[], "role":"assistant"}, "logprobs":null}], "created":1714586628, "model":"google-gemma-2b-it-q4", "systemFingerprint":"fp_ollama", "object":"chat.completion", "usage":{"completionTokens":248, "promptTokens":15, "totalTokens":263}}
```

server/internal/server/ollama_client.go
kkaneda: So line 88 calls `fn()` inside `scanner.Scan()`. Whenever `fn` is called we update `resp` here. Then we return `resp` at the end of this function. 

Is the assumption here that we only need to return the last `resp` as RPC response?

server/internal/server/ollama_client.go
kkaneda: I see. I guess this works as we also use the same message for gRPC gateway.

server/internal/server/ollama_client.go
```diff
@@ -1,72 +1,133 @@
...
-	return client, nil
 }
 
-func handleChatRequest(ctx context.Context, req *v1.CreateChatCompletionRequest) (*v1.ChatCompletion, error) {
-	client, err := newClient()
+const maxBufferSize = 512 * 1000
+
+func (c *client) stream(ctx context.Context, method, path string, data any, fn func([]byte) error) error {
+	var buf *bytes.Buffer
+	if data != nil {
```
kkaneda, line 42: Slightly prefer removing this `nil` check as we always expect a request in this context.

server/internal/server/ollama_client.go
```diff
@@ -1,72 +1,133 @@
...
+
+		bts := scanner.Bytes()
+		if err := json.Unmarshal(bts, &errorResponse); err != nil {
+			return fmt.Errorf("unmarshal: %w", err)
+		}
 
-	ollamaReq := &api.ChatRequest{
-		Model:    req.Model,
-		Messages: msgs,
+		if errorResponse.Error != "" {
```
kkaneda, line 80: Slightly prefer

```go
if e := errorResponse.Error; e != "" {
 return fmt.Errorf(e)
}
```

server/internal/server/ollama_client.go
```diff
@@ -1,72 +1,133 @@
...
+			return fmt.Errorf(errorResponse.Error)
+		}
+
+		if response.StatusCode >= http.StatusBadRequest {
+			return status.Errorf(codes.Code(response.StatusCode), "Status: %s: %s", response.Status, errorResponse.Error)
```
kkaneda, line 85: `errorResponse.Error` should be always `""`? If so, we don't need to print?

server/internal/config/config.go
```diff
@@ -40,6 +42,9 @@ func (c *Config) Validate() error {
 	if c.HTTPPort <= 0 {
 		return fmt.Errorf("httpPort must be greater than 0")
 	}
+	if c.MonitoringPort <= 0 {
```
kkaneda, line 45: Could you also check if `OllamaServerAddr` is not empty?

server/internal/server/server.go
```diff
@@ -14,15 +14,18 @@ import (
 )
 
 // New creates a server.
-func New() *S {
-	return &S{}
+func New(c *config.Config) *S {
```
kkaneda, line 17: How about just padding `OllamaServerAddr`?

server/internal/server/ollama_client.go
kkaneda: any thought?

server/internal/server/ollama_client.go
kkaneda: yeah i'm not sure if gRPC streaming works as it might not be compatible with the OpenAI API. We might need to just implement HTTP handlers without gRPC.

kkaneda (approved):
LG!

PR #47, opened 2024-05-01
setup inference-manager-server deployment

guangrui-cloudnatix:


deployments/server/templates/ingress.yaml
```diff
@@ -0,0 +1,39 @@
...
+metadata:
+  name: {{ include "inference-manager-server.fullname" . }}-grpc
+  labels:
+    {{- include "inference-manager-server.labels" . | nindent 4 }}
+spec:
+  ingressClassName: {{ .Values.global.ingress.ingressClassName }}
+  rules:
+  - http:
+      paths:
+      - path: /v1/chat/completions
```
kkaneda, line 12: The gRPC requires a different ingress definition. Let's remove this ingress if we don't need now.

deployments/server/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: {{ include "inference-manager-server.fullname" . }}-http
+  labels:
+    {{- include "inference-manager-server.labels" . | nindent 4 }}
+spec:
+  selector:
+    matchLabels:
+      {{- include "inference-manager-server.labels" . | nindent 6 }}
+  endpoints:
+  - port: monitoring
```
kkaneda, line 12: Can you add a return code at the end of the file?

deployments/server/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
```
kkaneda, line 2: Let's avoid adding this or this needs to be able to be disable via ahelm value. This doesn't work if `ServiceMonitor` CRD is not defined.

deployments/server/values.yaml
```diff
@@ -0,0 +1,20 @@
...
+# The following default values work if ollama server runs in the same namespace.
+ollamaServerAddr: llm-operator-inference-manager-engine-ollama:8080
+
+replicaCount: 1
+image:
+  repository: public.ecr.aws/v8n3t7y5/llm-operator/inference-manager-server
+  pullPolicy: IfNotPresent
+  version:
+
+ingress:
```
kkaneda, line 18: Remove this?

kkaneda:
Btw, if the plan is to put this in front of `inference-manager-engine`, we need to remove the ingress from `inference-manager-engine`?

guangrui-cloudnatix:
> Btw, if the plan is to put this in front of `inference-manager-engine`, we need to remove the ingress from `inference-manager-engine`?

yes, I think so

deployments/server/templates/deployment.yaml
```diff
@@ -0,0 +1,47 @@
...
+          containerPort: {{ .Values.httpPort }}
+          protocol: TCP
+        - name: grpc
+          containerPort: {{ .Values.grpcPort }}
+          protocol: TCP
+        - name: monitoring
+          containerPort: {{ .Values.monitoringPort }}
+          protocol: TCP
+        env:
+        - name: OLLAMA_HOST
```
kkaneda, line 38: Do we still need this?

PR #50, opened 2024-05-02
setup keda scaledobject for engine

guangrui-cloudnatix:


deployments/engine/templates/scaledobject.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: keda.sh/v1alpha1
```
kkaneda, line 1: Can you add a helm variable here and create this resource when enabled? The argument is same as `ServiceMonitor`. We expect Keda should be installed, but it's nice to have an option to disable it.

deployments/engine/templates/scaledobject.yaml
```diff
@@ -0,0 +1,26 @@
...
+kind: ScaledObject
+metadata:
+  name: {{ include "inference-manager-engine.fullname" . }}-scaledobject
```
kkaneda, line 4: Let's remove `-scaledobject`.

kkaneda (approved):
nice!

deployments/engine/values.yaml
```diff
@@ -31,3 +31,8 @@ ingress:
   ingressClassName:
 
 version:
+
+scaledobject:
+  minReplicaCount: 1
+  maxReplicaCount: 2
+  prometheusServerAddr: http://prometheus-server.monitoring.svc.cluster.local
```
kkaneda, line 38: I might later move this definition to https://github.com/llm-operator/llm-operator/blob/main/hack/llm-operator-values.yaml

deployments/engine/templates/scaledobject.yaml
```diff
@@ -0,0 +1,26 @@
...
+  # uncomment this if it is desired to scale to zero.
+  # idleReplicaCount: 0
+  triggers:
+    - type: prometheus
+      metadata:
+        serverAddress: {{ .Values.scaledobject.prometheusServerAddr }}
+        metricName: gpu-util
+        metricType: Value
+        query: |-
+          avg(avg_over_time(DCGM_FI_DEV_GPU_UTIL[1m]))
```
kkaneda, line 25: Question: Did you allocate GPU to the pod when running this test?

guangrui-cloudnatix: Yes, updated deployment spec too

deployments/engine/templates/deployment.yaml
```diff
@@ -44,6 +49,12 @@ spec:
             secretKeyRef:
               name: {{ .name }}
               key: {{ .secretAccessKeyKey }}
+        - name: PATH
```
kkaneda, line 52: Again I'm not sure if we need this.  Have you tried without this?

deployments/engine/templates/deployment.yaml
```diff
@@ -54,3 +65,7 @@ spec:
       - name: config
         configMap:
           name: {{ include "inference-manager-engine.fullname" . }}
+      tolerations:
```
kkaneda, line 68: I'm not sure if we need this.

deployments/engine/templates/deployment.yaml
kkaneda: Let me investigate.

kkaneda: - I don't think we need to set `$PATH` since we don't have the directory inside the container.
- Verified that `LD_LIBRARY_PATH`, etc. are set by the container runtime.

Will send a PR.

PR #55, opened 2024-05-02
Remove ingress from engine

kkaneda:
Requests will be routed to inference-manager-server first.


kkaneda:
I'm not too sure if we really need to have two deployments, but that's something we can revisit later.

Ladicle (approved):

PR #64, opened 2024-05-08
allow inference-manager-server to access k8s object

guangrui-cloudnatix:


deployments/server/templates/configmap.yaml
```diff
@@ -10,4 +10,5 @@ data:
     httpPort: {{ .Values.httpPort }}
     monitoringPort: {{ .Values.monitoringPort }}
     ollamaServerAddr: {{ .Values.ollamaServerAddr }}
+    inferenceEngineServerAddr: {{ .Values.inferenceEngineServerAddr }}
```
kkaneda, line 13: This one is not set? Also is this the internal gRPC endpoint or not?

deployments/server/templates/rbac.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
kkaneda, line 2: I think this can be just `Role` (not `ClusterRole`)?

server/internal/server/k8sclient.go
```diff
@@ -0,0 +1,44 @@
...
+	return &k, nil
+}
+
+// CoreClientset returns the Core clientset.
+func (k *K8sClient) CoreClientset() kubernetes.Interface {
+	return k.coreClientset
+}
+
+// GetEndPoints get a Service.
+func (k *K8sClient) GetEndPoints(ctx context.Context, name, namespace string) (*apiv1.Endpoints, error) {
```
kkaneda, line 41: not sure if endpoints are good ones, but we can just get pod ips. cc @Ladicle

kkaneda: Also not sure this will come eventually, but I think we need to watch instead of listing periodically.

guangrui-cloudnatix: Removed.

deployments/server/templates/configmap.yaml
guangrui-cloudnatix: Removed.

deployments/server/values.yaml
```diff
@@ -20,3 +20,7 @@ enableServiceMonitor: false
 
 service:
   annotations:
+
+serviceAccount:
+  create: true
+  name:
```
kkaneda, line 26:
```suggestion
  serviceAccountName: inference-manager-engine
```

kkaneda (approved):
LG!

deployments/server/values.yaml
guangrui-cloudnatix: wondering why we don't want `llm-operator-` prefix for this serviceaccount, and it is for `inference-manager-server` instead of `inference-manager-engine`?
```
kubectl get serviceaccounts -n llm-operator
NAME                                    SECRETS   AGE
default                                 0         14d
llm-operator-inference-manager-server   0         156m
llm-operator-job-manager-dispatcher     0         14d
llm-operator-job-manager-server         0         14d
```

kkaneda: My bad. It should be `inference-managa-server`. Also it's better to just remove this line.

Then https://github.com/llm-operator/inference-manager/blob/main/deployments/engine/templates/_helpers.tpl#L59 sets the service account name.

As for the `llm-operator-` prefix, Helm adds it since the chart name (eg.., `job-manager-dispatcher`) is not as the release name we specify in `helm install`. 

https://stackoverflow.com/questions/66841290/helm-adds-release-name-to-dependent-charts

This is not particularly intentional and we can fix if we want.

PR #76, opened 2024-05-16
dynamically load models per request

guangrui-cloudnatix:


kkaneda (commented):
Nice! Still looking.

server/internal/server/completions.go
```diff
@@ -40,10 +40,22 @@ func (s *S) CreateChatCompletion(
 	// TODO(kenji): Check if the specified model is available
 	// for the tenant.
 
+	if createReq.Model == "" {
+		http.Error(w, "Model is required", http.StatusInternalServerError)
```
kkaneda, line 44:
```suggestion
		http.Error(w, "Model is required", http.StatusBadRequest)
```

server/internal/server/route.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	for _, v := range found {
+		if v == ip {
+			return
+		}
+	}
+	r.m[model] = append(found, ip)
+}
+
+func (r *route) deleteRoute(model model) error {
```
kkaneda, line 50: unused?

server/internal/server/server.go
```diff
@@ -101,3 +126,144 @@ func (s *S) Run(ctx context.Context, port int, authConfig config.AuthConfig) err
 func (s *S) Stop() {
 	s.srv.Stop()
 }
+
+// RunRoute updates the routes periodically and upon pod events.
+func (s *S) RunRoute(ctx context.Context, errCh chan error) error {
+	k8sClient, err := k8s.NewClient()
```
kkaneda, line 132: Can't we define a `Router` struct and move the logic there? We can also have a separate package (e.g., `route` or `router`).

server/internal/server/server.go
```diff
@@ -101,3 +126,144 @@ func (s *S) Run(ctx context.Context, port int, authConfig config.AuthConfig) err
...
+func (s *S) getDestination(ctx context.Context, modelID string) (string, error) {
+	routes := s.m.getRoute(model{id: modelID})
+	if len(routes) != 0 {
+		// TODO(guangrui): handle load balance.
+		return routes[0], nil
+	}
+	return s.pullModel(ctx, modelID)
+}
+
+func (s *S) pullModel(ctx context.Context, modelID string) (string, error) {
```
kkaneda, line 248: The return value of this function wasn't clear to me. Might make sense to move line 249-253 to `getDestination`.

server/internal/server/server.go
kkaneda: Can't we also have some unit test?

server/internal/server/route.go
```diff
@@ -0,0 +1,132 @@
...
+func (r *route) replace(nr *route) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	r.m = nr.m
+	r.engines = nr.engines
+}
+
+// findLeastLoadedEngine finds the engine with the least number of loaded models.
+func (r *route) findLeastLoadedEngine() (string, error) {
```
kkaneda, line 95: need unit test?

server/internal/server/pod.go
```diff
@@ -0,0 +1,75 @@
+package server
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	"github.com/llm-operator/inference-manager/pkg/k8s"
```
kkaneda, line 8: Slightly prefer moving this package under `server/internal/k8s` as `engine` doesn't use. We can create a separate PR.

server/internal/config/config.go
```diff
@@ -54,8 +56,11 @@ func (c *Config) Validate() error {
 	if c.MonitoringPort <= 0 {
 		return fmt.Errorf("monitoringPort must be greater than 0")
 	}
-	if c.OllamaServerAddr == "" {
-		return fmt.Errorf("ollamaServerAddr must be set")
+	if c.OllamaPort <= 0 {
+		return fmt.Errorf("ollamaPort must be greater than 0")
+	}
+	if c.InferenceManagerEngineInternalGRPCPort <= 0 {
+		return fmt.Errorf("inferenceManagerEngineInternalGRPOCPort must be greater than 0")
```
kenji-cloudnatix, line 63:
```suggestion
		return fmt.Errorf("inferenceManagerEngineInternalGRPCPort must be greater than 0")
```

kkaneda:
@guangrui-cloudnatix , please let me know if there are unclear comments. There are certainly some personal preference, but wanted to keep modularity so that we can test easily.

We can also haver a pairing.

server/internal/server/server.go
guangrui-cloudnatix: Will add unit test later

server/internal/router/router.go
```diff
@@ -0,0 +1,184 @@
...
+)
+
+const tickPeriod = 60 * time.Second
+
+type model struct {
+	id string
+}
+
+// Router manages the routing table of models to engines.
+type Router struct {
```
kkaneda, line 24: This can be just `R` since the package name is `router`.

server/internal/router/router.go
```diff
@@ -0,0 +1,184 @@
...
+	// m is the route map.
+	m *route
```
kkaneda, line 26: nitpick: `r` might be better for the field name (or change `route` to `routeMap` as `route` sounds like a single route).

server/internal/router/router.go
```diff
@@ -0,0 +1,184 @@
...
+
+	for {
+		select {
+		case <-ticker.C:
+			if err := r.refreshRoutes(ctx, r.imeConfig.Namespace, engineLabels); err != nil {
+				return err
+			}
+			r.m.printRoute()
+		case <-stopper:
+			log.Printf("run done.\n")
```
kkaneda, line 86: Remove? Doesn't look very useful.

server/internal/router/router.go
```diff
@@ -0,0 +1,184 @@
...
+			return nil
+		case err := <-errCh:
+			log.Printf("run error: %s", err)
```
kkaneda, line 89: Same here.

server/internal/router/pod.go
```diff
@@ -0,0 +1,75 @@
...
+	"context"
+	"fmt"
+	"log"
+
+	"github.com/llm-operator/inference-manager/server/internal/k8s"
+	apiv1 "k8s.io/api/core/v1"
+)
+
+// PodEventHandler is the event handler for managing pod.
+type PodEventHandler struct {
```
kkaneda, line 13: unexpose?

server/internal/router/route.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+	}
+
+	if ip == "" {
+		return "", fmt.Errorf("no engine available")
+	}
+	return ip, nil
+}
+
+func (r *route) printRoute() {
```
kkaneda, line 126:
```suggestion
func (r *route) print() {
```

or `String()` and then let the caller print.

server/internal/router/router.go
```diff
@@ -0,0 +1,184 @@
...
+
+	engineLabels := map[string]string{
+		r.imeConfig.LabelKey: r.imeConfig.LabelValue,
+	}
+
+	// initialize the routes before starting the informer.
+	if err := r.refreshRoutes(ctx, r.imeConfig.Namespace, engineLabels); err != nil {
+		return err
+	}
+	r.m.printRoute()
```
kkaneda, line 58: I think we don't need this since `refreshRoutes` print?

server/internal/router/router.go
```diff
@@ -0,0 +1,184 @@
...
+	return fmt.Sprintf("%s:%d", ip, r.imeConfig.OllamaPort), nil
+}
+
+func (r *Router) refreshRoutes(ctx context.Context, namespace string, labels map[string]string) error {
+	k8sClient, err := k8s.NewClient()
+	if err != nil {
+		return fmt.Errorf("create K8sClient: %v", err)
+	}
+
+	pods, err := k8sClient.ListPods(ctx, namespace, labels)
```
kkaneda, line 115: We can change this later, but I thought informer can track a list of pods.

server/internal/router/router.go
```diff
@@ -0,0 +1,184 @@
...
+	var modelIDs []string
+	for _, model := range resp.Models {
+		modelIDs = append(modelIDs, model.Id)
+	}
+	return modelIDs, nil
+}
+
+// pullModel pulls the model to the least loaded engine and returns the IP address of the engine.
+func (r *Router) pullModel(ctx context.Context, modelID string) (string, error) {
+	ip, err := r.m.findLeastLoadedEngine()
```
kkaneda, line 164: I still think it's better to move line 164-167 to the caller.

server/internal/router/router.go
```diff
@@ -0,0 +1,184 @@
...
+func NewRouter(c config.InferenceManagerEngineConfig) *Router {
+	return &Router{
+		m:         newRoute(),
+		imeConfig: c,
+	}
+}
+
+// Run updates the routes periodically and upon pod events.
+func (r *Router) Run(ctx context.Context, errCh chan error) error {
+	k8sClient, err := k8s.NewClient()
```
kkaneda, line 41: I think we want to initialize the k8s client or informer outside of this class. Then we can inject a fake k8s client.

server/internal/router/router.go
```diff
@@ -0,0 +1,184 @@
...
+
+// pullModel pulls the model to the least loaded engine and returns the IP address of the engine.
+func (r *Router) pullModel(ctx context.Context, modelID string) (string, error) {
+	ip, err := r.m.findLeastLoadedEngine()
+	if err != nil {
+		return "", err
+	}
+
+	dest := fmt.Sprintf("%s:%d", ip, r.imeConfig.InternalGRPCPort)
+	log.Printf("Pulling model %s to %s\n", modelID, dest)
```
kkaneda, line 170:
```suggestion
	log.Printf("Pulling model %s on %s\n", modelID, dest)
```

?

server/internal/router/router.go
```diff
@@ -0,0 +1,184 @@
...
+	if err != nil {
+		return "", err
+	}
+	ec := v1.NewInferenceEngineInternalServiceClient(conn)
+	if _, err = ec.PullModel(ctx, &v1.PullModelRequest{Id: modelID}); err != nil {
+		return "", err
+	}
+
+	r.m.addRoute(model{id: modelID}, ip)
+	log.Printf("Done pulling model %s to %s\n", modelID, dest)
```
kkaneda, line 181:
```suggestion
	log.Printf("Finished pulling model %s on %s\n", modelID, dest)
```

server/internal/router/router.go
```diff
@@ -0,0 +1,184 @@
...
+	conn, err := grpc.Dial(dest, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		return "", err
+	}
+	ec := v1.NewInferenceEngineInternalServiceClient(conn)
+	if _, err = ec.PullModel(ctx, &v1.PullModelRequest{Id: modelID}); err != nil {
+		return "", err
+	}
+
+	r.m.addRoute(model{id: modelID}, ip)
```
kkaneda, line 180: My preference is to move this to the caller.

server/internal/router/router.go
```diff
@@ -0,0 +1,184 @@
...
+// pullModel pulls the model to the least loaded engine and returns the IP address of the engine.
+func (r *Router) pullModel(ctx context.Context, modelID string) (string, error) {
+	ip, err := r.m.findLeastLoadedEngine()
+	if err != nil {
+		return "", err
+	}
+
+	dest := fmt.Sprintf("%s:%d", ip, r.imeConfig.InternalGRPCPort)
+	log.Printf("Pulling model %s to %s\n", modelID, dest)
+	conn, err := grpc.Dial(dest, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
kkaneda, line 171: We probably want to define a factory for this and inject to this struct so that we can fake/mock in the unit test.

server/internal/router/pod.go
```diff
@@ -0,0 +1,75 @@
...
+) *podEventHandler {
+	return &podEventHandler{
+		m:     m,
+		errCh: errCh,
+	}
+}
+
+// ProcessAdd processed the add event.
+func (h *podEventHandler) ProcessAdd(ctx context.Context, obj interface{}) {
+	// o, ok := obj.(runtime.Object)
```
kkaneda, line 33: remove?

server/internal/router/pod.go
```diff
@@ -0,0 +1,75 @@
...
+	pod, ok := obj.(*apiv1.Pod)
+	if !ok {
+		h.errCh <- fmt.Errorf("cast runtime.Object type")
+		return
+	}
+	log.Printf("Create a pod, ip is %s\n", pod.Status.PodIP)
```
kkaneda, line 39:
```suggestion
	log.Printf("Creating a pod, ip is %s\n", pod.Status.PodIP)
```

Or maybe "Create" is a bit odd here since server itself is not creating a pod. Maybe

```suggestion
	log.Printf("Registering a pod (IP: %s) in the routing table\n", pod.Status.PodIP)
```

server/internal/router/pod.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	oldPod, ok := oldObj.(*apiv1.Pod)
+	if !ok {
+		h.errCh <- fmt.Errorf("cast runtime.Object type")
+		return
+	}
+	if newPod.Status.PodIP == oldPod.Status.PodIP {
+		return
+	}
+	log.Printf("Update a pod, new IP is %s, old IP is %s\n", newPod.Status.PodIP, oldPod.Status.PodIP)
```
kkaneda, line 60:
```suggestion
	log.Printf("Updating a pod, new IP is %s, old IP is %s\n", newPod.Status.PodIP, oldPod.Status.PodIP)
```

server/internal/router/pod.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// ProcessDelete processes the delete event.
+func (h *podEventHandler) ProcessDelete(ctx context.Context, obj interface{}) {
+	pod, ok := obj.(*apiv1.Pod)
+	if !ok {
+		h.errCh <- fmt.Errorf("cast runtime.Object type")
+		return
+	}
+	log.Printf("Delete a pod, ip is %s\n", pod.Status.PodIP)
```
kkaneda, line 73:
```suggestion
	log.Printf("Deleting a pod, ip is %s\n", pod.Status.PodIP)
```

server/internal/router/pod.go
```diff
@@ -0,0 +1,75 @@
...
+	oldPod, ok := oldObj.(*apiv1.Pod)
+	if !ok {
+		h.errCh <- fmt.Errorf("cast runtime.Object type")
+		return
+	}
+	if newPod.Status.PodIP == oldPod.Status.PodIP {
+		return
+	}
+	log.Printf("Update a pod, new IP is %s, old IP is %s\n", newPod.Status.PodIP, oldPod.Status.PodIP)
+	if oldPod.Status.PodIP == "" {
```
kkaneda, line 61: The code here is correct, but it implicitly assumes that a pod IP only changes from an empty value to a non-empty value once. It looks a bit weird that we won't do anything here if a pod IP changes from a non-empty value to a different non-empty value here.

How about changing the code to be more explicit?

```go
if oldPod.Status.PodIP != "" {
  h.errCh <- fmt.Errorf("unexpected ip change)
  return
}
h.m.addServer(newPod.Status.PodIP)
```

server/internal/k8s/k8sclient.go
```diff
@@ -11,12 +11,16 @@ import (
 )
 
 // Client is a wrapper for Kubernetes clientsets.
-type Client struct {
+type Client interface {
```
kkaneda, line 14: My preference is to define the interface when it is used as much as possible. This is not always better as it might cause redundant definitions, but https://go.dev/wiki/CodeReviewComments#interfaces has the following recommendation:

> Go interfaces generally belong in the package that uses values of the interface type, not the package that implements those values. The implementing package should return concrete (usually pointer or struct) types: that way, new methods can be added to implementations without requiring extensive refactoring.

For this case, I think we can define `podlister` interface where `Client` is consumed (= `refreshRoutes`).

server/internal/router/route.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	for _, v := range found {
+		if v == ip {
+			return
+		}
+	}
+	r.m[model] = append(found, ip)
+}
+
+func (r *routeMap) deleteRoute(model model) error {
```
kkaneda, line 50: unused? we need a unit test at leaste?

server/internal/router/router.go
```diff
@@ -0,0 +1,192 @@
...
+	return &R{
+		m:             newRouteMap(),
+		engineClients: make(map[string]v1.InferenceEngineInternalServiceClient),
+		imeConfig:     c,
+	}
+}
+
+// Run updates the routes periodically and upon pod events.
+func (r *R) Run(ctx context.Context, errCh chan error) error {
+	coreClient, err := k8s.NewCoreClient()
```
kkaneda, line 42: How about adding another parameter `New`  for this? As this is an immutable field that is consistently used in this struct, we can add a field to the struct and reuse, instead of passing to `refreshRoutes`.

server/internal/router/router.go
```diff
@@ -0,0 +1,192 @@
...
+	stopper := make(chan struct{})
+	defer close(stopper)
+	defer runtime.HandleCrash()
+
+	engineLabels := map[string]string{
+		r.imeConfig.LabelKey: r.imeConfig.LabelValue,
+	}
+
+	// initialize the routes before starting the informer.
+	if err := r.refreshRoutes(ctx, coreClient, r.imeConfig.Namespace, engineLabels); err != nil {
```
kkaneda, line 56: We don't need to pass `r.imeConfig.Namespace` to this function?

server/internal/router/router.go
```diff
@@ -0,0 +1,192 @@
...
+	coreClient, err := k8s.NewCoreClient()
+	if err != nil {
+		return fmt.Errorf("create K8sClient: %v", err)
+	}
+
+	stopper := make(chan struct{})
+	defer close(stopper)
+	defer runtime.HandleCrash()
+
+	engineLabels := map[string]string{
```
kkaneda, line 51: This is a personal preference, so please feel free to ignore, but it would be easier to read if we create `engineLabels` inside `refreshRoutes`. As this is also used in the informer, we can also define a helper function to construct `engineLabels`.

server/internal/router/router.go
kkaneda: still valid?

server/internal/router/router.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+	}
+}
+
+// GetEngineForModel returns the engine address (ip and port) for the given model.
+func (r *R) GetEngineForModel(ctx context.Context, modelID string) (string, error) {
+	routes := r.m.getRoute(model{id: modelID})
+	if len(routes) != 0 {
+		// TODO(guangrui): handle load balance.
+		return routes[0], nil
```
kkaneda, line 99: Question: Is this returning an IP or a pair of IP and a hostname?

server/internal/router/router.go
kkaneda: Actually it seems this is only useful at a startup time.

server/internal/router/router_test.go
```diff
@@ -0,0 +1,149 @@
...
+	r.engineClients = map[string]v1.InferenceEngineInternalServiceClient{
+		"1.2.3.4": engineClient,
+	}
+	r.m.addServer("1.2.3.4")
+
+	addr, err := r.GetEngineForModel(context.Background(), "model1")
+	assert.NoError(t, err)
+	assert.Equal(t, "1.2.3.4:8080", addr)
+
+	addr, err = r.GetEngineForModel(context.Background(), "model2")
```
kkaneda, line 147: `addr` is not used.

server/internal/router/route.go
```diff
@@ -0,0 +1,132 @@
...
+		m:       make(map[model][]string),
+		engines: make(map[string]struct{}),
+	}
+}
+
+func (r *routeMap) getRoute(model model) []string {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	return r.m[model]
```
kkaneda, line 30: I'm not sure about the concurrency guarantee there that a caller will access the slice outside of the mutex lock, and `addRoute` might be called.

kkaneda (commented):
Nice! Just some more comments.

server/internal/router/route.go
```diff
@@ -0,0 +1,132 @@
...
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	r.m = nr.m
+	r.engines = nr.engines
+}
+
+// findLeastLoadedEngine finds the engine with the least number of loaded models.
+func (r *routeMap) findLeastLoadedEngine() (string, error) {
+	modelsByIP := make(map[string][]model)
```
kenji-cloudnatix, line 96: Could you add a brief comment here?

```
// modelsIP is a list of models keyed by engine IPs. This includes engines that don't have any models
```

server/internal/router/route.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	r.mu.Unlock()
+
+	for k, v := range tmp {
+		for _, ip := range v {
+			modelsByIP[ip] = append(modelsByIP[ip], k)
+		}
+	}
+
+	min := 0
```
kenji-cloudnatix, line 111: Could you add a brief comment here?

```
// Find the engine that does not have the least number of models pulled.
```

server/internal/router/route.go
```diff
@@ -0,0 +1,132 @@
...
+	ip := ""
+	for k, v := range modelsByIP {
+		if len(v) < min || ip == "" {
+			min = len(v)
+			ip = k
+		}
+	}
+
+	if ip == "" {
```
kenji-cloudnatix, line 120: I think we should replace this to the following and put at the top of the function.

```go
if len(r.engines) == 0 {
  return "", fmt.Errorf("no engine available")
}
```

server/internal/router/router.go
guangrui-cloudnatix: `ListPods` need `namespace`

server/internal/router/router.go
guangrui-cloudnatix: Passing it to `refreshRoutes` is for making mock test easily, since `Infromer` need `coreClientSet`, but `refreshRoutes` need `listPods` only

server/internal/router/router.go
kkaneda: But, `refreshRoutes` is a receiver of `r`, so it can just get from `r.imeConfig.Namespace`?

server/internal/router/router.go
kkaneda: I'm not thinking that mocking might not be hard if we use the fake k8s client, but I could be wrong. We can revisit later.

kkaneda (approved):
LG!

PR #99, opened 2024-05-30
Check auth before pulling model to engine

guangrui-cloudnatix:


server/internal/server/completions.go
```diff
@@ -57,22 +57,24 @@ func (s *S) CreateChatCompletion(
...
+				http.Error(w, fmt.Sprintf("Model not found: %s", createReq.Model), http.StatusBadRequest)
+				return
+			}
+			http.Error(w, fmt.Sprintf("Failed to get model: %s", err), http.StatusInternalServerError)
 			return
 		}
-		http.Error(w, fmt.Sprintf("Failed to get model: %s", err), http.StatusInternalServerError)
+	}
+
+	dest, err := s.engineGetter.GetEngineForModel(req.Context(), createReq.Model)
```
guangrui-cloudnatix, line 75: Since it takes a long time to pull model to the engine, it is better to check if a user is allow to access the model before pulling it.

server/internal/server/completions.go
```diff
@@ -57,22 +57,24 @@ func (s *S) CreateChatCompletion(
...
-	}
-
 	// Check if the specified model is available
-	ctx := auth.CarryMetadataFromHTTPHeader(req.Context(), req.Header)
-	if _, err := s.modelClient.GetModel(ctx, &mv1.GetModelRequest{
-		Id: createReq.Model,
-	}); err != nil {
-		if status.Code(err) == codes.NotFound {
-			http.Error(w, fmt.Sprintf("Model not found: %s", createReq.Model), http.StatusBadRequest)
+	if s.enableAuth {
```
kkaneda, line 61: Hmm, this logic is not particularly tied to authentication? Is this related to your dev setting to avoid dep to model-manager-server?

kkaneda (approved):

server/internal/server/completions.go
kkaneda: I think we can merge, but I'm curious your dev setting. I was trying to see what's the best way to test inference-manager without less dependency.

server/internal/server/server.go
```diff
@@ -51,19 +51,23 @@ func New(
...
 		reqIntercepter: noopReqIntercepter{},
+		enableAuth:     enableAuth,
 	}
 }
 
 // S is a server.
 type S struct {
 	v1.UnimplementedChatServiceServer
 
+	enableAuth bool
```
kkaneda, line 69: How about changing `Run` so that it doesn't take `authConfig`?

server/internal/server/completions.go
guangrui-cloudnatix: My understanding was: `GetModel` fails if 

- model does not exist at all. `GetEngineForModel` returns error for this case too, so by-pass the check if `Auth` is disabled, no need to check twice?
- model exists but is not accessible for this user. `GetEngineForModel` does not check for this.

Not sure if any other cases missed?

kkaneda: Yeah, that's correct. I wasn't sure whether we need to have a special logic here. It seems to me that we need to this if we disable auth in `inference-manager-server`, but keeps the auth in `model-manager-server`?

kkaneda: Again we can merge the PR, but I was just curious..

PR #106, opened 2024-05-31
add metrics on number of active requests per model

guangrui-cloudnatix:


server/internal/server/completions.go
```diff
@@ -33,6 +33,10 @@ func (s *S) CreateChatCompletion(
 	st := time.Now()
 	defer func() {
 		s.metricsMonitor.ObserveCompletionLatency(createReq.Model, time.Since(st))
+		if createReq.Model == "" {
```
kkaneda, line 36: I think we don't need this check since we increment the counter after checking `createReq.Model` is not empty.

server/internal/monitoring/monitoring.go
```diff
@@ -10,17 +10,20 @@ const (
 	metricNamespace = "llm_operator"
 
 	metricsNameCompletionLatency = "inference_manager_serever_completion_latency"
+	metricsNameCompletionRequest = "inference_manager_serever_completion_request"
```
kkaneda, line 13: How about `inference_manager_server_completion_num_active_requests"`?

kkaneda (approved):
LG!

server/internal/server/completions.go
guangrui-cloudnatix: If the request returns in between line 43-64, the gauge with model ID "" will be negative?

kkaneda: Ah sorry, you're right, but I think we can still hit a negative case if `createReq.Model` is not empty but hit some other error (e.g., json marshal error).

How about just having a separate `defer` at line 68? Then it's clear.

kkaneda (approved):

PR #110, opened 2024-06-07
proto for vector store

guangrui-cloudnatix:
Define vector store management API following [OpenAPI API specification](https://platform.openai.com/docs/api-reference/vector-stores)

Not sure if vector store management should be in a new component.


guangrui-cloudnatix:
Move the proto def to `file-manager`

PR #123, opened 2024-06-28
Introduce queue abstraction

kkaneda:
No functional change, but introduce a queue when processing chat completion requests.

This is for up-coming change to support multi-cluster environment.

### DIFF
--- a/server/internal/infprocessor/infprocessor.go
+++ b/server/internal/infprocessor/infprocessor.go
@@ -18,7 +18,7 @@ const (
 	completionPath = "/v1/chat/completions"
 )
 
-// Task is an infeference task.
+// Task is an inference task.
 type Task struct {
 	Req    *v1.CreateChatCompletionRequest
 	Header http.Header
--- a/server/internal/infprocessor/infprocessor.go
+++ b/server/internal/infprocessor/infprocessor.go
@@ -1,0 +1,136 @@
+package infprocessor
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"log"
+	"net/http"
+	"net/url"
+
+	v1 "github.com/llm-operator/inference-manager/api/v1"
+	"github.com/llm-operator/inference-manager/common/pkg/models"
+)
+
+const (
+	taskQueueSize  = 100
+	completionPath = "/v1/chat/completions"
+)
+
+// Task is an infeference task.
+type Task struct {
+	Req    *v1.CreateChatCompletionRequest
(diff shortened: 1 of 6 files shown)


server/internal/infprocessor/infprocessor.go
```diff
@@ -0,0 +1,136 @@
...
+	v1 "github.com/llm-operator/inference-manager/api/v1"
+	"github.com/llm-operator/inference-manager/common/pkg/models"
+)
+
+const (
+	taskQueueSize  = 100
+	completionPath = "/v1/chat/completions"
+)
+
+// Task is an infeference task.
```
guangrui-cloudnatix, line 21: "inference"

server/internal/infprocessor/infprocessor.go
```diff
@@ -0,0 +1,136 @@
...
+type Task struct {
```
guangrui-cloudnatix, line 22: not sure, wondering if we need to preserve  `context` also?

server/internal/infprocessor/infprocessor.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// Run runs the processor.
+func (p *P) Run(ctx context.Context) error {
+	for {
+		t, err := p.queue.Dequeue(ctx)
+		if err != nil {
+			return err
+		}
+		p.runTask(ctx, t)
```
guangrui-cloudnatix, line 92: bypass `runTask` if t is nil?

guangrui-cloudnatix: also, it seems a busy-loop, not sure if it should sleep somewhere?

guangrui-cloudnatix (approved):

server/internal/infprocessor/infprocessor.go
kkaneda:
```suggestion
// Task is an inference task.
```

server/internal/infprocessor/infprocessor.go
kkaneda: Ah `t` should be non-nil when `err` is `nil`. Also Dequeue blocks when there is no task as the channel read is blocking by default.

server/internal/infprocessor/infprocessor.go
kkaneda: good point! not sure. let me add a TODO.
```suggestion
// TODO(kenji): Consider preserving the request context as well.
type Task struct {
```

PR #128, opened 2024-07-01
Add ID to the inference task

kkaneda:
To uniquely identity tasks between engine and server.


server/internal/server/completions.go
```diff
@@ -72,7 +73,12 @@ func (s *S) CreateChatCompletion(
 		return
 	}
 
+	taskID, err := id.GenerateID("inf_", 22)
```
Ladicle, line 76: just checking, most IDs are 24 in length, but not aligned here?

Ladicle (approved):

server/internal/server/completions.go
kkaneda: good call! Let me fix.

kkaneda:
```suggestion
	taskID, err := id.GenerateID("inf_", 24)
```

PR #129, opened 2024-07-01
Make engine register via gRPC call

kkaneda:
Engine makes a bi-directional gRPC call to server and register itself.

Later we will add a logic to fetch inference tasks via the same streaming RPC call.


Ladicle (approved):
LGTM

PR #140, opened 2024-07-03
Make engine pull a model when processing a task

kkaneda:


engine/internal/modelsyncer/syncer.go
```diff
@@ -82,12 +89,6 @@ func (s *S) PullModel(ctx context.Context, modelID string) error {
 }
 
 func (s *S) registerBaseModel(ctx context.Context, modelID string) error {
-	s.mu.Lock()
```
kkaneda, line 85: This change is no-op. Moving the logic to the callee.

Ladicle (approved):

PR #156, opened 2024-07-11
Check if a vector store is accessible from server

kkaneda:
Return a bad request error if a specified vector store is not found.


Ladicle (approved):
LG!

PR #171, opened 2024-07-23
Avoid a deadlock when a context is cannceled

kkaneda:
When the context is canceled from WaitForCompletion, we no longer read from the channels of the task, but the processor still attempts to write when it receives a message from the engine.

We need to spawn a gorountine to read messages whenever comes. We could also increase the channel buffer to 1 (and keep sufficient buffer for resp.Body), but just reading from channels seems simpler.

### DIFF
--- a/server/internal/infprocessor/infprocessor.go
+++ b/server/internal/infprocessor/infprocessor.go
@@ -3,6 +3,7 @@ package infprocessor
 import (
 	"context"
 	"fmt"
+	"io"
 	"log"
 	"net/http"
 	"sync"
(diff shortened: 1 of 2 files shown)


server/internal/infprocessor/infprocessor.go
```diff
@@ -33,18 +34,32 @@ type Task struct {
...
 		return resp, nil
 	case err := <-t.ErrCh:
-		log.Printf("Task failed: %s\n", err)
+		log.Printf("Task (ID: %q) failed: %s\n", t.ID, err)
 		return nil, err
 	}
 }
 
+func (t *Task) discardResp() {
+	select {
```
Ladicle, line 54: Question: does either respCh or errCh always return a value after a context is canceled?

kkaneda: Yes, that's correct! 

- If the engine is able to correctly process a request, the server receives a response and write to `respCh`.
- if some error happens (e.g., engine is disconnected), an error is written to `errCh`.

For example, if an engine is disconnected, we call `RemoveEngine`, which writes an error to all in-progress tasks.

https://github.com/llm-operator/inference-manager/blob/ebe48f7c31014a3e83c972f4049b602150c8aff1/server/internal/infprocessor/infprocessor.go#L192

PR #172, opened 2024-07-24
Be able to create/use a service account

kkaneda:
This is required when a user needs to a service account to give a S3 bucket access.

### DIFF
--- a/deployments/engine/templates/serviceaccount.yaml
+++ b/deployments/engine/templates/serviceaccount.yaml
@@ -1,0 +1,8 @@
+{{- if .Values.serviceAccount.create -}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "inference-manager-dispatcher.serviceAccountName" . }}
+  labels:
+    {{- include "inference-manager-dispatcher.labels" . | nindent 4 }}
+{{- end -}}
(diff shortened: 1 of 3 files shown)


deployments/engine/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,8 @@
+{{- if .Values.serviceAccount.create -}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "inference-manager-dispatcher.serviceAccountName" . }}
```
guangrui-cloudnatix, line 5: "inference-manager-engine.serviceAccountName"?

deployments/engine/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,8 @@
...
+  labels:
+    {{- include "inference-manager-dispatcher.labels" . | nindent 4 }}
```
guangrui-cloudnatix, line 7: "engine" vs "dispatcher"?

deployments/engine/templates/serviceaccount.yaml
kkaneda: ah, sorry, forgot to cc. :( https://github.com/llm-operator/inference-manager/pull/173/files fixed.

PR #190, opened 2024-08-01
Make the engine process tasks in parallel

kkaneda:


engine/internal/processor/processor.go
```diff
@@ -89,47 +89,79 @@ type P struct {
...
+			return fmt.Errorf("receive task: %s", err)
 		}
 
-		log.Printf("Started processing task: %s\n", resp.NewTask.Id)
-		if err := p.processTask(ctx, stream, resp.NewTask); err != nil {
-			return err
-		}
-		log.Printf("Completed task: %s\n", resp.NewTask.Id)
+		// Create a goroutine to process the task so that we can receive the
+		// next task. Ollama then might process requests in parallel.
```
Ladicle, line 156: question: no needs to limit how many tasks process in parallel?

Ladicle (approved):

engine/internal/processor/processor.go
kenji-cloudnatix: Good point! We might introduce the limit later, but decided not to do now. The reason is to keep the logic simple. If we want to limit the number of requests in parallel, `inference-manager-server` should do so that it can distribute requests to other `engine` instance. Also vLLM can have its queue, so we don't need to have yet another queue here.

PR #192, opened 2024-08-01
Guangrui/vllm build

guangrui-cloudnatix:


kkaneda (approved):
LG!

I'm also fine with just creating a separate Dockerfile. I'm not sure how easy to set this in the github workflow runner.

Instead of creating a different image repo, we could also use tag.

cc @guangrui-cloudnatix

PR #194, opened 2024-08-01
Remove the service from engine

kkaneda:
This is no longer used as all requests are fetched from the server.


Ladicle (approved):

guangrui-cloudnatix (commented):
`vector-store-manager` needs to access ollama/vllm to generate embedding, refer to https://github.com/llm-operator/vector-store-manager/blob/f33be46957032d81f84032784ff9ac7e80d7fb63/deployments/server/values.yaml#L36

How do we plan to handle such request? Add into `inference manager server` API, or revoke this change? In long term, we probably should support  `embedding` api in openai, e.g. https://api.openai.com/v1/embeddings.  @kkaneda @Ladicle

kkaneda:
Ah I'm sorry. I missed that. Let me revert. 

Yeah we can call the `inference-manager-server` API or maybe have a separate small engine.

kkaneda:
https://github.com/llm-operator/inference-manager/pull/208

PR #207, opened 2024-08-05
Define const for LLM kind

kkaneda:


engine/internal/llmkind/llmkind.go
```diff
@@ -0,0 +1,12 @@
+package llmkind
```
guangrui-cloudnatix, line 1: might want to move it to pkg, since `vector-store-manager` also will refer to ollama and vllm

guangrui-cloudnatix (approved):

engine/internal/llmkind/llmkind.go
kkaneda: sounds good!

PR #214, opened 2024-08-08
Configure securityContext, resources, and probes in the deployments

Ladicle:


deployments/engine/templates/deployment.yaml
```diff
@@ -69,6 +73,12 @@ spec:
         - name: config
           mountPath: /etc/config
           readOnly: true
+        - name: tmp
```
kkaneda, line 76: Could you add some comment to explain why this change is needed? My understanding is that this is to load the models in memory. If so, I think this should be fine, but we just need to be a bit careful to make sure if we have sufficient memory when multiple models are loaded (this only apply to Ollama, though).

kkaneda (approved):
LG!

deployments/engine/templates/deployment.yaml
kkaneda: Ah nvm. I misunderstood the behavior of `emptyDir`.

deployments/engine/templates/deployment.yaml
```diff
@@ -69,6 +73,12 @@ spec:
...
+          mountPath: /tmp
+        - name: tmp
+          mountPath: /.ollama
```
guangrui-cloudnatix, line 79: Can we rename this to something like `/.llm` or `/.engine`, since this should be applicable to the engine with `vLLM as well?

guangrui-cloudnatix (approved):

deployments/engine/templates/deployment.yaml
Ladicle: This name depends on ollama behavior, so if another engine writes something to the root filesystem, we need to add the path here.
ollama generates the key at startup in `/.ollama`, and this change allows ollama to write to it.

guangrui-cloudnatix: I see.  Thanks for the explanation. I'm not sure if vllm needs to write root filesystem.

PR #228, opened 2024-08-11
Do not crash when inference-manager-server is unavailable

kkaneda:
It can easily take more than several minutes to download & load large models. It is better to avoid a pod crash when the server is unreachable for a short period of time.

### DIFF
--- a/deployments/engine/templates/deployment.yaml
+++ b/deployments/engine/templates/deployment.yaml
@@ -36,6 +36,9 @@ spec:
         - name: llm
           containerPort: {{ .Values.llmPort }}
           protocol: TCP
+        - name: health
+          containerPort: {{ .Values.healthPort }}
+          protocol: TCP
         env:
         {{- with .Values.global.awsSecret }}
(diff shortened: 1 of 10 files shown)


kkaneda:
@Ladicle , i'll merge this so that this change is available Monday, but PR review appreciated. thanks!

deployments/engine/templates/deployment.yaml
```diff
@@ -72,6 +75,8 @@ spec:
           mountPath: /tmp
         - name: tmp
           mountPath: /.ollama
+        readinessProbe:
```
Ladicle, line 78: I'm curious why a readiness probe with one replica count is used. Is it for detecting errors in somewhere alerting systems and handling failures manually?

kkaneda: Yeah that's the main reason. In particular when the initial installation fails, we would like to let end users know.

Also this helps when inference-manager-engine is get deployed. An existing pod is kept until a new one becomes ready.

PR #232, opened 2024-08-12
Increase the default keep alive to 96 hours

kkaneda:
This can be longer. Given that loading a large model like llama3.1 70B is slow, we would like to avoid end users from hitting the model loading as much as possible.


kkaneda:
I have changed my mind and decided to keep the models longer (96 hours).

PR #233, opened 2024-08-12
Change back to flush more often

kkaneda:


kkaneda:
I removed flush to have a better response for auto code completion, but decided to add back to make the chat response more smooth.

PR #255, opened 2024-08-21
Be able to configure a model context length

kkaneda:


kkaneda:
Let me go ahead and merge to test with the overall helm chart.

PR #265, opened 2024-08-26
Avoid using illegal characters in statefulset names

kkaneda:


engine/internal/runtime/client.go
```diff
@@ -275,3 +276,13 @@ func (c *commonClient) deployRuntime(
 	}
 	return nil
 }
+
+func statefulSetName(runtime, modelID string) string {
```
Ladicle, line 280: I feel `resourceName` would be nicer since this name is also used for SVC and PVC.

Ladicle (approved):

engine/internal/runtime/client.go
kkaneda: done!

PR #266, opened 2024-08-26
Support vLLM in puller

kkaneda:


engine/internal/runtime/client.go
```diff
@@ -178,6 +178,7 @@ func (c *commonClient) deployRuntime(
 		"--index=$(INDEX)",
 		"--runtime=" + c.Name,
 		"--model-id=" + modelID,
+		"--vllm-model-dir=" + modelDir,
```
Ladicle, line 181: I'd personally prefer not to add a vLLM-specific flag to the common client. I'd rather use an extra argument or environment for this. I think it would also be nice to create a public function in the runtime package that might be useful for the `isModelRegistered` function too.

Ladicle (approved):

engine/internal/runtime/client.go
kkaneda: sounds good!

PR #270, opened 2024-08-27
Use a fix dir for ollama payload

kkaneda:
Ollama creaets a payload in a temporary directory by default, and a new temporary directory is created whenever Ollama restarts. This is a problem when a persistent volume is mounted. To avoid this, we set the directory to a fixed path.


aya-cloudnatix (approved):
LG!

PR #272, opened 2024-08-27
Be able to use consistent format for model IDs

kkaneda:
Model IDs we use in model-manager-loader contains "/". Allow end users to have the same format in inference-manager-engine.


Ladicle (approved):

kkaneda:
I need to update one more place - resource allocation per model. Will update later.

PR #286, opened 2024-08-29
Log a request when --v is set to 1

kkaneda:


kkaneda:
@Ladicle , I haven't used this library before. Please let me know if you see any weird usage pattern. Thanks!

engine/internal/processor/processor.go
```diff
@@ -253,12 +263,13 @@ func (p *P) processTask(
 		return fmt.Errorf("build request: %s", err)
 	}
 
-	log.Printf("Sending request to the LLM server: %s\n", req.URL)
+	log.Info(fmt.Sprintf("Sending request to the LLM server: %s", req.URL))
```
Ladicle, line 266: Since it's a structured logging library, it's common practice to split the message into key-value pairs for variables.
```suggestion
	log.Info("Sending request to the LLM server", "url", req.URL)
```

engine/internal/processor/processor.go
```diff
@@ -268,20 +279,21 @@ func (p *P) processTask(
...
 
+	log.Info(fmt.Sprintf("Received an initial response from the LLM server: status=%q", resp.Status))
+
 	defer func() {
 		_ = resp.Body.Close()
 	}()
 
 	if resp.StatusCode < http.StatusOK || resp.StatusCode >= http.StatusBadRequest {
-		log.Printf("Received an error response from llm: statusCode=%d, status=%q\n", resp.StatusCode, resp.Status)
+		log.Info(fmt.Sprintf("Received an error response from the LLM server: statusCode=%d, status=%q", resp.StatusCode, resp.Status))
```
Ladicle, line 291:
```suggestion
		log.Info("Received an error response from the LLM server", "statusCode", resp.StatusCode, "status", resp.Status)
```

Ladicle (approved):

engine/internal/processor/processor.go
kkaneda: got it!

PR #287, opened 2024-08-30
Support the AWQ quantization in vLLM execution

kkaneda:
This is one of the changes required to support AWQ.

We need to change model-manager and modelsyncher to make this work.

### DIFF
--- a/engine/internal/vllm/models.go
+++ b/engine/internal/vllm/models.go
@@ -14,7 +14,7 @@ func IsAWQQuantizedModel(modelName string) bool {
 // ModelFilePath returns the file path of the model.
 func ModelFilePath(modelDir, modelName string) string {
 	if IsAWQQuantizedModel(modelName) {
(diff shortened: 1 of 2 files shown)


Ladicle (approved):

engine/internal/vllm/models.go
```diff
@@ -6,8 +6,19 @@ import (
...
 
+// IsAWQQuantizedModel returns true if the model name is an AWQ quantized model.
+func IsAWQQuantizedModel(modelName string) bool {
+	return strings.HasSuffix(modelName, "-awq")
+}
+
 // ModelFilePath returns the file path of the model.
 func ModelFilePath(modelDir, modelName string) string {
+	if IsAWQQuantizedModel(modelName) {
+		// VLLM requires an entire directory with the HuggingFace file structure.
```
kkaneda, line 17:
```suggestion
		// vLLM requires the entire directory with the HuggingFace file structure.
```

PR #305, opened 2024-09-05
Improve scaling stability

Ladicle:


engine/internal/metrics/metrics.go
```diff
@@ -2,35 +2,161 @@ package metrics
...
 	defer c.mu.RUnlock()
-	return c.metrics[modelID]
+	wb, ok := c.metrics[modelID]
+	if !ok {
+		return 0
+	}
+	return wb.average(time.Now())
+}
+
+// percision is the granularity of the window bucket.
```
kenji-cloudnatix, line 53:
```suggestion
// precision is the granularity of the window bucket.
```

engine/internal/autoscaler/autoscaler.go
```diff
@@ -113,16 +114,27 @@ type scaler struct {
...
-func (s *scaler) start(log logr.Logger, stopCh <-chan struct{}, period time.Duration) {
+func (s *scaler) start(log logr.Logger, stopCh <-chan struct{}, initialDelay, period time.Duration) {
 	ctx, cancel := context.WithCancel(context.Background())
 	s.cancel = cancel
 
 	log = log.WithName("scaler").WithValues("modelID", s.modelID)
 	ctx = ctrl.LoggerInto(ctx, log)
 
-	ticker := time.NewTicker(period)
+	runOnce := func() {
```
kenji-cloudnatix, line 124: Really personal preference, but how about `scale` for the function name?

engine/internal/autoscaler/autoscaler.go
```diff
@@ -157,21 +163,25 @@ func (s *scaler) scale(ctx context.Context) error {
 		return client.IgnoreNotFound(err)
 	}
 
-	totalObserved := s.metricsClient.Get(s.modelID)
-	observedPerPods := totalObserved / float64(sts.Status.Replicas)
+	// TODO(aya): Stop scaling when there are too many not-ready pods
 
+	metricsVal := s.metricsClient.Get(s.modelID)
+	replicas := math.Min(float64(sts.Status.ReadyReplicas), float64(sts.Status.Replicas))
```
kenji-cloudnatix, line 169: might make sense to define our own `min` and `max` and avoid conversion to `float64`.

engine/internal/autoscaler/autoscaler.go
```diff
@@ -182,12 +192,24 @@ func (s *scaler) scale(ctx context.Context) error {
 	} else {
 		s.lastTransitToZero = time.Time{}
 	}
-	return s.scaleTo(ctx, &sts, cappedReplicas)
+
+	specReplicas := float64(ptr.Deref(sts.Spec.Replicas, 0))
```
kenji-cloudnatix, line 196: Could you add a brief comment on the logic between line 196-207?

engine/internal/config/config.go
```diff
@@ -191,6 +197,9 @@ type ScalingConfig struct {
 
 	MaxReplicas int `yaml:"maxReplicas"`
 	MinReplicas int `yaml:"minReplicas"`
+
+	MaxScaleUpRate   float64 `yaml:"maxScaleUpRate"`
```
kenji-cloudnatix, line 201: Could you add a brief comment on these fields? Some example case that explains the behavior would be useful.

engine/internal/autoscaler/autoscaler.go
```diff
@@ -182,12 +192,24 @@ func (s *scaler) scale(ctx context.Context) error {
...
+
+	if newReplicas < specReplicas {
+		maxScaleDown := math.Floor(float64(replicas) * s.config.MaxScaleDownRate)
```
kenji-cloudnatix, line 199: Checking my understanding. If`MaxScaleDownRate` is 0.2, and the current replica size is 10, we scale down the stateful set to 2 replicas when it is most aggressive. 

When the number of replica becomes 2, `maxScaleDown` is `floor(2*0.2)= 0`. If the min replica is 0, the target replica size becomes 0.

kenji-cloudnatix (commented):
Nice! Still looking.

Ladicle:
rebase onto main

PR #306, opened 2024-09-05
download s3 object concurrently

guangrui-cloudnatix:
to be tested e2e


engine/internal/huggingface/download_test.go
```diff
@@ -64,7 +65,7 @@ func TestDownloadModelFiles(t *testing.T) {
 				_ = os.RemoveAll(dir)
 			}()
 
-			err = DownloadModelFiles(tc.s3Client, "/src", dir)
+			err = DownloadModelFiles(context.TODO(), tc.s3Client, "/src", dir)
```
kkaneda, line 68:
```suggestion
			err = DownloadModelFiles(context.Background(), tc.s3Client, "/src", dir)
```

engine/internal/s3/client.go
```diff
@@ -1,42 +1,50 @@
...
 
-// Download downloads an object from S3.
-func (c *Client) Download(w io.WriterAt, key string) error {
-	downloader := s3manager.NewDownloaderWithClient(c.svc)
-	_, err := downloader.Download(w, &s3.GetObjectInput{
+// Download uses a download manager to download an object from a bucket.
+// The download manager gets the data in parts and writes them to a buffer until all of
+// the data has been downloaded.
+func (c *Client) Download(ctx context.Context, w io.WriterAt, key string) error {
+	var partMiBs int64 = 10
```
kkaneda, line 43:
```suggestion
	const partMiBs int64 = 10
```

Also the part size can be bigger? Not sure though.

kkaneda (approved):
LG!

engine/internal/s3/client.go
guangrui-cloudnatix: Can be larger. It basically depends on if we want concurrent downloading for a small or medium object, e.g. 50MB or 100MB object. Any suggestions on what default value to use?

kkaneda: So 

- `model.gguf` is the largest as it contains the entire model. Can be >10GiB
- safetensors files is severeal GiB.
- Other json files are few KB or MB. 

Maybe use 128MB as a chunk size that gives sufficient concurrency for large file download.

PR #308, opened 2024-09-06
Promote 'alpha' as a default command

kkaneda:
- Rename `alpha` to `run` 
- Rename original `run` to `run_monolithic`


kkaneda:
This will change the default behavior.

PR #312, opened 2024-09-09
Use distroless for a base image for inference-manager-engine

kkaneda:
Still build an image with Ollama base so that we can fall back to the previous behavior, but change the default image to not include Ollama anymore.


.github/workflows/ci-post-merge.yml
```diff
@@ -71,9 +72,10 @@ jobs:
         file: ./build/server/Dockerfile
         push: true
         platforms: linux/amd64,linux/arm64
+        target: base-ollama
```
kkaneda, line 75: This is a bug where I changed the server. Will fix.

PR #318, opened 2024-09-10
Decopule puller from Ollama

kkaneda:
With this change, 'puller' will simply downloads the model and create a modelfile, but it doesn't actually register the model in Ollama.

We create another init container that runs 'ollama serve' and 'ollama create'.

This will also help us support additional runtimes in the future as the puller code can be simply reused.


kkaneda:
I'd like to merge this change now, but please let us know if you have any review comments. thanks!

PR #319, opened 2024-09-10
Make Register & Unregister no-op when autoscaling is disabled

kkaneda:
Manager calls Register() and Unregister(). We shouldn't call them when autoscaling is disabled.

### DIFF
--- a/engine/cmd/run.go
+++ b/engine/cmd/run.go
@@ -19,6 +19,7 @@ import (
 	"github.com/spf13/cobra"
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc"
+	"k8s.io/apimachinery/pkg/types"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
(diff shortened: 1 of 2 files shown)


engine/cmd/run.go
```diff
@@ -81,11 +82,14 @@ func run(ctx context.Context, c *config.Config, ns string, lv int) error {
 	}
 
 	mClient := metrics.NewClient(c.Autoscaler.MetricsWindow)
-	scaler := autoscaler.NewMultiAutoscaler(mgr.GetClient(), mClient, c.Autoscaler)
+	var scaler runtime.ScalerRegisterer
 	if c.Autoscaler.Enable {
+		scaler := autoscaler.NewMultiAutoscaler(mgr.GetClient(), mClient, c.Autoscaler)
```
Ladicle, line 87: seems autoscaler inserted into the local val

kkaneda: my bad! let me fix.

PR #321, opened 2024-09-10
Add leader election option

Ladicle:


engine/internal/autoscaler/autoscaler.go
```diff
@@ -86,7 +94,10 @@ func (m *MultiAutoscaler) Register(modelID string, target types.NamespacedName)
 		scaleToZeroGracePeriod: m.config.ScaleToZeroGracePeriod,
 	}
 	m.scalers[target] = s
-	s.start(m.logger, m.stopCh, m.config.InitialDelay, m.config.SyncPeriod)
+	go func() {
+		<-m.startCh
```
kkaneda, line 98: Could you explain why this is needed? Couldn't understand why we need to wait for `Start` to be called.

Ladicle: `Start()` will start after the pod is elected as a leader if leader election is enabled, and this logic is for syncing the lifecycle with multi-scaler and scaler.

Ladicle:
resolve conflicts

PR #351, opened 2024-09-16
Gracefully handle task requests

Ladicle:


engine/internal/processor/processor.go
```diff
@@ -174,32 +159,48 @@ func (p *P) NeedLeaderElection() bool {
...
-	}()
+		defer func() { _ = stream.CloseSend() }()
 
-	go func() {
-		if err := p.processTasks(ctx, stream); err != nil {
-			errCh <- fmt.Errorf("process tasks: %s", err)
+		eg, ctx := errgroup.WithContext(ctx)
+		eg.Go(func() error { return p.sendEngineStatusPeriodically(ctx, stream) })
+		eg.Go(func() error { return p.processTasks(ctx, stream) })
+		return eg.Wait()
```
kkaneda, line 186: Supposet that sendEngineStatusPeriodically() returns an error. `eg.Wait` does not return the error until `processTasks` completes IIUC. Does this work when `inference-manager-server` becomes unavailable for a short period of time?

engine/internal/processor/processor.go
```diff
@@ -285,9 +313,26 @@ func (p *P) processTask(
 	if err := p.modelSyncer.PullModel(ctx, taskModel(t)); err != nil {
 		return fmt.Errorf("pull model: %s", err)
 	}
-	// TODO(aya): Consider how to handle the case where the runtime is scaled down before the request is sent.
 
-	req, err := p.buildRequest(ctx, t)
+	done := make(chan string)
```
kkaneda, line 317: I guess `atomic.Bool` is another option? No opinion on either way, though.

engine/internal/processor/processor.go
```diff
@@ -174,32 +159,48 @@ func (p *P) NeedLeaderElection() bool {
...
-
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
 
-	errCh := make(chan error, 2)
-	go func() {
-		if err := p.sendEngineStatusPeriodically(ctx, stream); err != nil {
-			errCh <- fmt.Errorf("send engine status: %s", err)
+	run := func() error {
+		streamCtx, streamCancel := context.WithCancel(context.Background())
```
kkaneda, line 172: Out of of scope of this PR, but can we make `inference-manager-engine` stop receiving new tasks when grace period starts?

kkaneda (approved):
LG!

engine/internal/processor/processor.go
Ladicle: Yes, in that case, `stream.Recv()` fails, and `processTasks` returns an error.

engine/internal/processor/processor.go
kenji-cloudnatix: got it!

PR #355, opened 2024-09-18
Retry task processing on temporary engine failure

Ladicle:


server/internal/infprocessor/pipe.go
```diff
@@ -28,6 +28,9 @@ func (c pipeReadWriteCloser) Close() error {
 
 // closeWrite closes the write pipe.
 func (c pipeReadWriteCloser) closeWrite() error {
+	if c.PipeWriter == nil {
```
kkaneda, line 31: How about changing `task.bodyWriter` to ` *pipeReadWriteCloser` and add a nil-check before calling `t.bodyWriter.closeWrite()`?

```
diff --git a/server/internal/infprocessor/infprocessor.go b/server/internal/infprocessor/infprocessor.go
index cff8edb..51ba0fe 100644
--- a/server/internal/infprocessor/infprocessor.go
+++ b/server/internal/infprocessor/infprocessor.go
@@ -38,7 +38,7 @@ type task struct {
 	Header http.Header
 
 	setResult  func(*result)
-	bodyWriter pipeReadWriteCloser
+	bodyWriter
 
 	EngineID string
 
@@ -392,8 +392,10 @@ func (p *P) RemoveEngine(engineID string, clusterInfo *auth.ClusterInfo) {
 		p.logger.Info("Canceled task", "reason", "engine removed", "engine", engineID, "task", t.ID)
 		delete(p.inProgressTasksByID, t.ID)
 		t.setResult(&result{err: fmt.Errorf("engine %s is removed", engineID), retry: true})
-		if err := t.bodyWriter.closeWrite(); err != nil {
-			p.logger.Error(err, "Failed to close the body writer when engine removed", "engine", t.EngineID, "task", t.ID)
+		if t.bodyWriter != nil {
+			if err := t.bodyWriter.closeWrite(); err != nil {
+				p.logger.Error(err, "Failed to close the body writer when engine removed", "engine", t.EngineID, "task", t.ID)
+			}
 		}
 	}
 
diff --git a/server/internal/infprocessor/pipe.go b/server/internal/infprocessor/pipe.go
index 86090f4..6482407 100644
--- a/server/internal/infprocessor/pipe.go
+++ b/server/internal/infprocessor/pipe.go
@@ -2,9 +2,9 @@ package infprocessor
 
 import "io"
 
-func newPipeReadWriteCloser() pipeReadWriteCloser {
+func newPipeReadWriteCloser() *pipeReadWriteCloser {
 	pr, pw := io.Pipe()
-	return pipeReadWriteCloser{
+	return &pipeReadWriteCloser{
 		PipeReader: pr,
 		PipeWriter: pw,
 	}
@@ -28,9 +28,6 @@ func (c pipeReadWriteCloser) Close() error {
 
 // closeWrite closes the write pipe.
 func (c pipeReadWriteCloser) closeWrite() error {
-	if c.PipeWriter == nil {
-		return nil
-	}
 	if err := c.PipeWriter.Close(); err != nil {
 		return err
 	}
```

server/internal/infprocessor/infprocessor.go
```diff
@@ -20,90 +20,46 @@ const (
...
-		Header:    header,
-		RespCh:    make(chan *http.Response),
-		ErrCh:     make(chan error),
-		CreatedAt: time.Now(),
-		logger:    logger.WithValues("id", taskID),
-	}, nil
+type result struct {
+	resp  *http.Response
+	err   error
+	retry bool
```
kkaneda, line 26: This not used now?

server/internal/infprocessor/infprocessor.go
```diff
@@ -293,16 +230,108 @@ func (p *P) findLeastLoadedEngine(engineIDs []string) string {
 	var minTasks int
 	var leastLoaded string
 	for _, engineID := range engineIDs {
+		if p.unavailableEngines[engineID] {
```
kkaneda, line 233: Can't this be redundant as we already remove engines from `engineRouter` and `engineRouter.GetEnginesForModel` doesn't return unavailable engines?

server/internal/infprocessor/infprocessor.go
```diff
@@ -317,22 +346,37 @@ func (p *P) AddOrUpdateEngineStatus(
...
 	if s := engineStatus.SyncStatus; s != nil {
 		e.inProgressModelIDs = s.InProgressModelIds
 	}
+	curUnavailable := p.unavailableEngines[engineStatus.EngineId]
+	if curUnavailable && !engineStatus.Unavailable {
+		delete(p.unavailableEngines, engineStatus.EngineId)
+		log.Info("Marked the engine as available")
+	} else if !curUnavailable && engineStatus.Unavailable {
+		p.unavailableEngines[engineStatus.EngineId] = true
+		p.engineRouter.DeleteEngine(engineStatus.EngineId, clusterInfo.TenantID)
```
kkaneda, line 370: Redundant as we call at line 376?

server/internal/infprocessor/infprocessor.go
```diff
@@ -293,16 +230,108 @@ func (p *P) findLeastLoadedEngine(engineIDs []string) string {
...
+	embeddingReq *v1.CreateEmbeddingRequest,
+	header http.Header,
+	logger logr.Logger,
+) (*http.Response, error) {
+	taskID, err := id.GenerateID("inf_", 24)
+	if err != nil {
+		return nil, fmt.Errorf("generate task ID: %s", err)
+	}
+	log := logger.WithValues("id", taskID)
+	done := make(chan *result, 1)
```
kkaneda, line 278: Could you explain why the buffer size needs to be 1?

Ladicle: This is because it guarantees the retrieval of one result and discards further results. The ordinary, done channel gets one result, but an additional error occurs when the engine is removed. Previously, the goroutine was used to send the error to the channel to prevent blocking in the removed engine. However, the goroutine would leak if the engine was removed during task processing second and later response since no one received the channel after getting the initial response.

kenji-cloudnatix: I see. Can you leave the comment? Thanks!

PR #356, opened 2024-09-20
Add runtime updater

Ladicle:


engine/internal/runtime/manager.go
```diff
@@ -182,12 +182,15 @@ func (m *Manager) PullModel(ctx context.Context, modelID string) error {
...
 		}
-		nn, err := client.DeployRuntime(ctx, modelID)
+		sts, err := client.DeployRuntime(ctx, modelID, false)
 		if err != nil {
 			cleanup()
 			return err
 		}
-		m.autoscaler.Register(modelID, nn)
+		m.autoscaler.Register(modelID, types.NamespacedName{Name: sts.Name, Namespace: sts.Namespace})
+		if sts.Status.ReadyReplicas > 0 {
```
kkaneda, line 191: Could you explain why this change is needed? Can we just rely on `Reconcile`? Also `Reconcile` checks if the runtime is reachable before marking it ready, and it seems the logic here is inconsistent with that.

engine/internal/runtime/updater.go
```diff
@@ -0,0 +1,74 @@
...
+		client.MatchingLabels{
+			"app.kubernetes.io/name":       "runtime",
+			"app.kubernetes.io/created-by": managerName,
+		}); err != nil {
+		return fmt.Errorf("failed to list runtimes: %s", err)
+	}
+
+	for _, sts := range stsList.Items {
+		modelID := sts.GetAnnotations()[modelAnnotationKey]
+		client, err := u.rtClientFactory.New(modelID)
```
kkaneda, line 61: Question: Does this handle a case where we change the runtime (e.g., from ollama to vllm)? We don't need to cover such a case now, but just wanted to understand.

kkaneda (approved):
LG!

engine/internal/runtime/manager.go
Ladicle: This is because if there is already a ready runtime, processing will proceed without waiting for the leader-election component to update the k8s resources. Since the number of ready replicas will never be more than 1 in the response when it is created, this is a process for registering an existing runtime, and it's the same as L235.

engine/internal/runtime/updater.go
Ladicle: No, this only supports runtime updates but does not support deletion.

engine/internal/runtime/manager.go
kkaneda: If there is a ready runtime, shouldn't we hit this code path (a line 160 and line 162  return true)? Anyway, if I'm wrong, can you leave a comment? It's getting difficult to track/imagine all the control flow, so comments would help. (Otherwise it will be difficult to refactor later as we will forget why the code is here.)

kkaneda: I think I now understand why we need this. 

Maybe we can refactor later and call `addRuntime` here as well. Anyway LG!

kkaneda (approved):

kkaneda:
thanks!

PR #358, opened 2024-09-20
Define helper function reconcile()

kkaneda:
Call reconcile() from PullModel. This should consolidate the model readiness check we have.

### DIFF
--- a/engine/internal/runtime/manager.go
+++ b/engine/internal/runtime/manager.go
@@ -64,7 +64,7 @@ type runtime struct {
 	waitCh chan struct{}
 }
 
-func (m *Manager) addRuntime(modelID string, sts appsv1.StatefulSet) (bool, error) {
+func (m *Manager) addRuntime(modelID string, sts *appsv1.StatefulSet) (bool, error) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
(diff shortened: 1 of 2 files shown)


engine/internal/runtime/manager.go
```diff
@@ -187,15 +187,20 @@ func (m *Manager) PullModel(ctx context.Context, modelID string) error {
...
-			m.markRuntimeReady(modelID, client.GetAddress(sts.Name))
+
+		// If this is called before the first cache sync of the reconciler
+		// is complete, the existing statefulset(STS) for the runtime is not
+		// registered in the manager's managed runtime map.
+		//
+		// Run reconcile() here so that if the runtime's STS is already ready,
+		// mark the runtime as ready without waiting for the reconciler
+		// (leader-election component) to process it.
+		if err := m.reconcile(ctx, sts); err != nil {
```
Ladicle, line 198: If we use reconcile here, this function could update the resource before the leader election if the resource has a deletion timestamp. Since the reconciler and processor's startup timing is different to prevent resource updates, it would be better to process only the runtime registration part without sharing the reconciliation.

kkaneda: I see. So wanted to avoid taking mutating actions from a non-leader. Let me withdraw and think.

PR #389, opened 2024-10-04
enable lora-based fine tuning model in vllm

guangrui-cloudnatix:


Ladicle (approved):
LG!

engine/internal/runtime/vllm_client.go
```diff
@@ -166,7 +225,8 @@ func chatTemplate(modelID string) (string, error) {
 	switch {
 	case strings.HasPrefix(modelID, "meta-llama-Meta-Llama-3.1-"),
 		strings.HasPrefix(modelID, "TinyLlama-TinyLlama-1.1B-Chat-v1.0"),
-		strings.HasPrefix(modelID, "mattshumer-Reflection-Llama-3.1-70B"):
+		strings.HasPrefix(modelID, "mattshumer-Reflection-Llama-3.1-70B"),
+		strings.Contains(modelID, "TinyLlama-1.1B"):
```
kkaneda, line 229: This one includes line 227. Can't we restrict the check or remove line 227?

engine/internal/s3/client.go
```diff
@@ -56,3 +56,31 @@ func (c *Client) Download(ctx context.Context, w io.WriterAt, key string) error
...
+	f func(page *s3.ListObjectsV2Output, lastPage bool) bool,
+) error {
+	const maxKeys = int32(1000)
+	p := s3.NewListObjectsV2Paginator(
+		c.svc,
+		&s3.ListObjectsV2Input{
+			Bucket: aws.String(c.bucket),
+			Prefix: aws.String(prefix),
+		},
+		func(o *s3.ListObjectsV2PaginatorOptions) {
```
kkaneda, line 74: This looks unnecessary since the default limit is 1000.

https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/s3#ListObjectsV2PaginatorOptions

> By default, the action returns up to 1,000 key name

kkaneda (commented):
Nice! Still looking.

engine/internal/modeldownloader/huggingface/download.go
```diff
@@ -8,18 +8,40 @@ import (
...
 const siFilename = "model.safetensors.index.json"
 
 type s3Client interface {
 	Download(ctx context.Context, f io.WriterAt, path string) error
+	ListObjectsPages(ctx context.Context, prefix string, f func(page *s3.ListObjectsV2Output, lastPage bool) bool) error
 }
 
 // DownloadModelFiles downloads model files from S3.
 func DownloadModelFiles(ctx context.Context, s3Client s3Client, srcS3Path string, destDir string) error {
+	keys, err := listFiles(ctx, s3Client, srcS3Path)
```
kkaneda, line 26: Currently some users don't have the `ListBucket` permission, so this change can break it. Can we gracefully handle the case?

engine/internal/modeldownloader/huggingface/download.go
```diff
@@ -8,18 +8,40 @@ import (
...
+	if err != nil {
+		return fmt.Errorf("list files: %s", err)
+	}
+	if len(keys) == 0 {
+		return fmt.Errorf("no objects found under %q", srcS3Path)
+	}
+
+	found := false
```
kkaneda, line 34: Can you also add some comment on the rationale behind this change? If we only look for some specific objects, can get call GetObject?

kkaneda (approved):
LGTM!

PR #412, opened 2024-10-17
feat(runtime): support affinity and per-pod volume settings

Ladicle:


deployments/engine/values.yaml
```diff
@@ -34,6 +34,10 @@ runtime:
     # and TensortRT-LLM.
     triton: nvcr.io/nvidia/tritonserver:24.09-trtllm-python-py3
   imagePullPolicy: IfNotPresent
+  # If you want to controle pod and volume availability zone and topologies,
```
kkaneda, line 37:
```suggestion
  # If you want to control pod and volume availability zone and topologies,
```

engine/internal/config/config.go
```diff
@@ -222,9 +236,17 @@ type Resources struct {
 
 // PersistentVolume is the persistent volume configuration.
 type PersistentVolume struct {
-	StorageClassName string `yaml:"storageClassName"`
-	Size             string `yaml:"size"`
-	AccessMode       string `yaml:"accessMode"`
+	ShareWithReplicas bool   `yaml:"shareWithReplicas"`
```
kkaneda, line 239: Could you add a brief comment to this field?

engine/internal/runtime/client.go
```diff
@@ -137,36 +138,60 @@ func (c *commonClient) deployRuntime(
...
+	if vol := resConf.Volume; vol != nil {
+		pvcName := volName
+		if resConf.Volume.ShareWithReplicas {
+			pvcName = name
+			volumes = append(volumes, corev1apply.Volume().WithName(volName).
+				WithPersistentVolumeClaim(
+					corev1apply.PersistentVolumeClaimVolumeSource().
+						WithClaimName(name)))
+		} else {
+			forcePull = true
```
kkaneda, line 158: Could you add a comment here? Couldn't follow why we don't need to append to `volumes`.

engine/internal/runtime/client.go
```diff
@@ -413,3 +435,111 @@ func resourceName(runtime, modelID string) string {
...
+	lslAC := func(lsl *metav1.LabelSelector) *metav1apply.LabelSelectorApplyConfiguration {
+		ac := metav1apply.LabelSelector()
+		if len(lsl.MatchLabels) > 0 {
+			ac = ac.WithMatchLabels(lsl.MatchLabels)
+		}
+		for _, lse := range lsl.MatchExpressions {
+			ac = ac.WithMatchExpressions(metav1apply.LabelSelectorRequirement().
+				WithKey(lse.Key).
+				WithOperator(metav1.LabelSelectorOperator(lse.Operator)).
+				WithValues(lse.Values...))
```
kkaneda, line 473: Question. Should we call `WithValues` only when `lse.Values` is not empty?

engine/internal/runtime/client.go
```diff
@@ -413,3 +435,111 @@ func resourceName(runtime, modelID string) string {
 
 	return fmt.Sprintf("%s-%s", runtime, m)
 }
+
+func buildAffinityApplyConfig(affinity *corev1.Affinity) *corev1apply.AffinityApplyConfiguration {
```
kkaneda, line 439: How about adding a simple test for this?

kkaneda (approved):
LG!

PR #439, opened 2024-10-23
feat(server): define internal server

kkaneda:
Add an interface and a stub impl. This is used to exchange tasks among servers.


Ladicle (approved):
LG!

PR #441, opened 2024-10-23
refactor(server): refactor the logic for processing tasks results

kkaneda:
Move the logic from ProcessTaskResult to sendTask.

This is needed as an upcoming change requires a function that just forwards task results to a remote server without any additional processing.

### DIFF
--- a/server/internal/infprocessor/infprocessor.go
+++ b/server/internal/infprocessor/infprocessor.go
@@ -205,11 +205,7 @@ func (p *P) scheduleTask(ctx context.Context, t *task) error {
 			Header:                          header,
 		},
 	}); err != nil {
-		close(t.resultCh)
-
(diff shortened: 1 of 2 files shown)


Ladicle (approved):
Just a comment, but in Go, it's recommended to only pass a channel for reading to prevent panic, and to limit the goroutines that have ownership (initialize/write/close) of the channel. When I modified infprocessor.go last time, there was a case where writing after closing caused a panic, so I think keeping channels as centralized as possible can help prevent bugs.

kkaneda:
ack! it's been difficult to reason the correctness of the code. I'll revisit.

`resultCh` is currently the problematic one. It is created by `sendTask` and write to the channel happens from functions invoked from the gRPC handlers. The channel is closed from random places when the task completes successfully or with an error.

PR #443, opened 2024-10-24
feat(server): implement task exchanger

kkaneda:


kkaneda:
Will add unit tests later, but this PR is ready for review. This is working fine in a Kind cluster.

server/cmd/run.go
```diff
@@ -196,5 +258,10 @@ func run(ctx context.Context, c *config.Config, lv int) error {
 		errCh <- adminSrv.Run(c.AdminPort)
 	}()
 
+	go func() {
+		log.Info("Starting manager")
+		errCh <- mgr.Start(signals.SetupSignalHandler())
```
Ladicle, line 263: I might be wrong, but only the manager will be shut down first, and the server won't stop until the manager has finished. However, wouldn’t skipping the graceful shutdown lead to fewer request drops?

Ladicle (approved):

kenji-cloudnatix (commented):
Thanks for the review, @Ladicle !

PR #444, opened 2024-10-24
feat(server): add SendAndProcessTask to infprocessor.P

kkaneda:
The function is used to process a task forwarded from other server instance.

Make another refactoring and clean up the logic for updating `inProgressTasksByID`. It is now consolidated in `enqueueAndProcessTask`.


Ladicle (approved):
LGTM

PR #447, opened 2024-10-24
fix(common): fix a race condition in NewTestLogger

kkaneda:
Call stdr.SetVerbosity(8) from init() to avoid a test failure like below:

https://github.com/llmariner/inference-manager/actions/runs/11504696862/job/32024755965

### DIFF
--- a/common/pkg/test/log.go
+++ b/common/pkg/test/log.go
@@ -13,7 +13,6 @@ import (
 
 // NewTestLogger returns a logger that writes to the test log.
 func NewTestLogger(t *testing.T) logr.Logger {
-	stdr.SetVerbosity(8)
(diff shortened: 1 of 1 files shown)


Ladicle (approved):
sorry, this is my bad, I've added a custom logger here. But, `logr ("github.com/go-logr/logr/testr")` lib already has a testing library, so we can fix here like below:
```go
// NewTestLogger returns a logger that writes to the test log.
func NewTestLogger(t *testing.T) logr.Logger {
	return testr.NewWithOptions(t, testr.Options{Verbosity: 8})
}
```

kkaneda:
Oh I see. Thanks! Updated!

PR #461, opened 2024-11-01
feat(api): support multiple content types in chat completion request

guangrui-cloudnatix:


api/v1/inference_server.proto
```diff
@@ -22,10 +29,21 @@ message CreateChatCompletionRequest {
     }
 
     // The type of content varies. For eample, it is string or array for UserMesasge.
+    // Refer to https://platform.openai.com/docs/api-reference/chat/create#chat-create-messages
+    message Content {
+      ContentType type = 1;
+      string text = 2;
+      string image_url = 3;
+      string input_audio = 4;
```
kkaneda, line 37: This should be a proto message that contains `data` and `format`?

image

api/v1/inference_server.proto
```diff
@@ -22,10 +29,21 @@ message CreateChatCompletionRequest {
...
+    }
+    repeated Content contents = 5;
```
kkaneda, line 39: Could you clarify that this is not following the OpenAI API spec? The OpenAI API only has the `content` field whose type is string or array. Unfortunately we cannot encode that in the proto easily.

kkaneda (approved):

PR #463, opened 2024-11-04
feat(chart): productionize helm chart

Ladicle:


deployments/server/templates/configmap.yaml
```diff
@@ -28,7 +27,7 @@ data:
       {{- toYaml .Values.global.usageSender | nindent 6 }}
     kubernetesManager:
       enableLeaderElection: {{ .Values.kubernetesManager.enableLeaderElection }}
-      leaderElectionID: {{ include "job-manager-dispatcher.fullname" . }}
+      leaderElectionID: {{ include "inference-manager-server.fullname" . }}
```
kkaneda, line 30: thanks!

deployments/server/templates/deployment.yaml
```diff
@@ -55,8 +55,18 @@ spec:
           containerPort: {{ .Values.adminPort }}
           protocol: TCP
         - name: km-metrics
-          containerPort: {{ .Values.kubernetesManagerMetricsPort }}
+          containerPort: {{ include "inference-manager-server.metricsPort" . }}
```
kkaneda, line 58: I might be missing, but this is not defined?

kkaneda (approved):
LG!

deployments/server/templates/deployment.yaml
Ladicle: It's defined in the helper template.

kkaneda: ah i see. nice!

PR #465, opened 2024-11-05
feat(server): add an option to enable/disable dynamic model loading

kkaneda:


deployments/server/values.yaml
```diff
@@ -148,6 +148,10 @@ certificate:
     # The default value is Issuer (i.e. a locally namespaced Issuer)
     kind: ClusterIssuer
 
+requestRouting:
+  # Specify whether dynamic on-demand model loading is enabled.
```
Ladicle, line 152:
```suggestion
  # Specify whether dynamic on-demand model loading is enabled.
  # If set to false, the request will fail immediately if the specified
  # model has not yet been loaded on any cluster.
```

Ladicle (approved):
LG

It seems the values schema is not generated.

deployments/server/values.yaml
kkaneda: done.

kkaneda:
@Ladicle , how did you generate the values schema? I followed https://www.arthurkoziel.com/validate-helm-chart-values-with-json-schemas/

Ladicle:
make generate-chart-schema (or make lint)

kkaneda:
thx!

PR #470, opened 2024-11-07
fix(values.yaml): fix typos

takeshi-cloudnatix:


Ladicle (approved):
Could you run `make lint` to update the schema as well?

takeshi-cloudnatix:
> Could you run `make lint` to update the schema as well?

Yeah, both `helm-lint` and `generate-chart-schema` make no change. The schema file contains only an entry for `service.annotations`. None for the intermediate node `service`.

Ladicle:
Thanks! LG

PR #477, opened 2024-11-11
ci: add scripts for testing local inference server/engine

Ladicle:


hack/overwrite-llma-chart-for-test.py
```diff
@@ -0,0 +1,43 @@
+import yaml
+import sys
+
```
kkaneda, line 3: Could you add one line description to this file?

kkaneda (approved):
Thanks, @Ladicle ! This is very useful!

guangrui-cloudnatix (approved):

PR #480, opened 2024-11-13
fix(engine): fix leader-election and runtime deletion handling

Ladicle:
Fix leader-election settings and event handling for runtime deletion to handle request correctly with multiple replicas.

### DIFF
--- a/engine/internal/runtime/manager.go
+++ b/engine/internal/runtime/manager.go
@@ -9,11 +9,13 @@ import (
 
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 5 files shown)


engine/internal/runtime/manager.go
```diff
@@ -291,5 +305,8 @@ func (m *Manager) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&appsv1.StatefulSet{}, builder.WithPredicates(filterByAnno)).
 		WithLogConstructor(constructer).
+		// To share the runtime deletion event, disable the leader election
+		// for this controller if the processor disables it.
```
kkaneda, line 309: Slightly prefer "if the processor disables the leader election" to clarify what "it" refers to.

engine/internal/runtime/manager.go
```diff
@@ -75,14 +79,22 @@ func (m *Manager) addRuntime(modelID string, sts appsv1.StatefulSet) (bool, erro
...
 
-func (m *Manager) deleteRuntime(modelID string) {
+func (m *Manager) deleteRuntime(name string) {
+	var modelID string
+	for id, r := range m.runtimes {
+		if r.name == name {
+			modelID = id
+			break
+		}
+	}
```
kkaneda, line 96: How about returning immediately or return an error if `modelID` is empty?

kkaneda (approved):
LG!

PR #492, opened 2024-11-15
feat(api): support both array and string for the "content" field

kkaneda:
- Add the new "content" field (= different proto field ID). The type of the field is an array.
- If the "content" contains a string value in the request JSON, convert it to an array before proto unmarshalling.
- Rename the current "content" field to the "legacy_content", but continue to set the value. This is requierd for backward compatibility.


api/v1/inference_server.proto
```diff
@@ -28,25 +21,29 @@ message CreateChatCompletionRequest {
...
-      ContentType type = 1;
+      // Refer to https://platform.openai.com/docs/api-reference/chat/create#chat-create-messages
+      message ImageUrl {
+	string url = 1;
+	string detail = 2;
+      }
+      message InputAudio {
+	string data = 1;
+	string format = 2;
+      }
```
Ladicle, line 33: nits: it seems like the indent is off.

server/internal/server/legacy_completions.go
```diff
@@ -222,8 +222,8 @@ func toCreateChatCompletionRequest(req *v1.CreateCompletionRequest) *v1.CreateCh
 	return &v1.CreateChatCompletionRequest{
 		Messages: []*v1.CreateChatCompletionRequest_Message{
 			{
-				Content: req.Prompt,
-				Role:    "user",
+				//Content: []string{req.Prompt},
```
Ladicle, line 225: maybe
```suggestion
```

server/internal/server/marshal_test.go
```diff
@@ -0,0 +1,46 @@
...
+	assert.NoError(t, err)
+
+	/*
+
+
+		var req v1.CreateChatCompletionRequest
+		err := json.Unmarshal([]byte(body), &req)
+		assert.NoError(t, err)
+		assert.Equal(t, "You are a helpful assistant.", req.Messages[0].Content)
+	*/
```
Ladicle, line 45:
```suggestion
```

triton-proxy/internal/server/triton.go
```diff
@@ -18,7 +18,10 @@ func buildEnsembleGenerateRequest(req *v1.CreateChatCompletionRequest) *ensemble
 	// this to a template.
 	input := "<|begin_of_text|>"
 	for _, m := range req.Messages {
-		input += fmt.Sprintf("<|start_header_id|> %s <|end_header_id|>\n %s '\n<|eot_id|>\n", m.Role, m.Content)
+		for _, c := range m.Content {
+			// TODO(kenji): Support other types.
+			input += fmt.Sprintf("<|start_header_id|> %s <|end_header_id|>\n %s '\n<|eot_id|>\n", m.Role, c.Text)
```
Ladicle, line 23: This has not been changed in this PR, but is the `'` before `eot_id` a template syntax?

Ladicle (approved):
LGTM

server/internal/server/marshal_test.go
kkaneda: my bad. let me delete this file.

triton-proxy/internal/server/triton.go
kkaneda: nice catch! this looks like a bug. let me create a separate PR to fix.

api/v1/inference_server.proto
kkaneda: ah confusing, but I've moved these messages inside `Content` to follow the pattern in this file.

guangrui-cloudnatix (approved):

server/internal/server/completions.go
```diff
@@ -67,12 +67,26 @@ func (s *S) CreateChatCompletion(
...
+		return
+	}
+
 	// TODO(kenji): Use runtime.JSONPb from github.com/grpc-ecosystem/grpc-gateway/v2.
 	// That one correctly handles the JSON field names of the snake case.
 	if err := json.Unmarshal(reqBody, &createReq); err != nil {
 		httpError(w, err.Error(), http.StatusInternalServerError, &usage)
 		return
 	}
+	// Set the LegacyContext field for the backward compatibility.
```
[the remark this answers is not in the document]
kkaneda:
```suggestion
	// Set the LegacyContent field for the backward compatibility.
```

kkaneda:
Tested locally with Ollama.  Confirmed that the server with this change works with an older version of engine.

PR #497, opened 2024-11-21
feat: support request-based rate-limiter

Ladicle:


deployments/server/values.yaml
```diff
@@ -158,6 +158,47 @@ requestRouting:
   # model has not yet been loaded on any cluster.
   enableDynamicModelLoading: true
 
+# Settings for rate-limiting a inference request.
```
kenji-cloudnatix, line 161:
```suggestion
# Settings for rate-limiting inference requests.
```

Or 

```suggestion
# Settings for rate-limiting an inference request.
```

deployments/server/values.yaml
```diff
@@ -158,6 +158,47 @@ requestRouting:
...
+    # The address of the Redis server.
+    address: redis-master.redis:6379
+
+    # Optional username of the Redis server.
+    # +docs:property
+    # username: ""
+
+    # The secret name that is mounted to the pod.
+    # +docs:property
+    # secretName: redis
```
kenji-cloudnatix, line 195: not too sure about the convention, but shouldn't we comment this out if this is not an optional field? Or should we make the field optional?

server/internal/rate/config.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// Rate is allowed tokens for the period.
+	Rate int `yaml:"rate"`
+	// Period is the time period for the rate.
+	Period time.Duration `yaml:"period"`
+	// Burst is the maximum burst capacity.
+	Burst int `yaml:"burst"`
+}
+
+// rate returns the token emission intervalSec in seconds.
```
kenji-cloudnatix, line 30:
```suggestion
// intervalSec returns the token emission intervalSec in seconds.
```

server/internal/rate/gcra_ratelimit.lua
```diff
@@ -0,0 +1,42 @@
...
+local burst_offset = tonumber(ARGV[3])
+
+-- Get current time as seconds and microseconds from Redis
+-- Convert to float for precision; adjust the epoch to be relative
+-- to Jan 1, 2020 00:00:00 GMT to avoid floating point problems.
+local jan_1_2020 = 1577836800
+local rtime = redis.call("TIME")
+local now = (rtime[1] - jan_1_2020) + (rtime[2] / 1000000)
+
+local tat = tonumber(redis.call("GET", key)) or now
```
kenji-cloudnatix, line 15: Could you add some comment here? Something like `-- Calculate the theoretical arrival time (TAT) of the next token`

server/internal/rate/store.go
```diff
@@ -0,0 +1,190 @@
...
+	Take(ctx context.Context, key string, cost int) (*Result, error)
+}
+
+// Result is the result of a Take call.
+type Result struct {
+	// Allowed is true if the token is available.
+	Allowed bool
+	// Limit is the maximum number of tokens.
+	Limit int
+	// Remaining is the number of remaining token.
```
kenji-cloudnatix, line 30:
```suggestion
	// Remaining is the number of remaining tokens.
```

server/internal/rate/store.go
```diff
@@ -0,0 +1,190 @@
...
+	Remaining int
+	// ResetAfter is the duration until the token is available.
```
kenji-cloudnatix, line 32:
```suggestion
	// RetryAfter is the duration until the next token becomes available.
```

kenji-cloudnatix (approved):
Looks great!

PR #517, opened 2024-12-12
feat(engine)!: add KEDA integration for vLLM

Ladicle:


engine/internal/autoscaler/keda.go
```diff
@@ -0,0 +1,129 @@
...
+	s.logger.Info("Registering scaler", "modelID", modelID, "target", target.GetName())
+	if err := s.deployScaledObject(ctx, modelID, target); err != nil {
+		return fmt.Errorf("failed to deploy ScaledObject: %s", err)
+	}
+	return nil
+}
+
+// Unregister unregisters the scaler for the given runtime.
+func (s *KedaScaler) Unregister(target types.NamespacedName) {
+	// do nothing. ScaledObject has a owner reference so it will be deleted by GC.
```
kkaneda, line 72:
```suggestion
	// do nothing. ScaledObject has an owner reference so it will be deleted by GC.
```

engine/internal/autoscaler/keda.go
```diff
@@ -0,0 +1,129 @@
...
+// Start starts the multi-autoscaler.
+func (s *KedaScaler) Start(ctx context.Context) error {
+	s.logger.Info("Starting keda-scaler")
+	// TODO(aya): support KEDA external scaler
+	return nil
+}
+
+// NeedLeaderElection implements LeaderElectionRunnable and always returns true.
+func (s *KedaScaler) NeedLeaderElection() bool {
+	return true
```
kkaneda, line 58: Just curious, why does this need to be `true`? No need to change, but just wondered there is no ham if two engines attempt to create the same ScaleObject (and one of them succeeds).

kkaneda (approved):
LG!

engine/internal/autoscaler/keda.go
Ladicle: Ah, yes, we don't need to enable this for the current scaler. I originally thought I would provide the inference metrics as a KEDA external scaler, which I'm working on now, but if we only send an activation request through the external scaler pusher, this may be unnecessary. I’ll turn off this for now.

Ladicle, line 58:
```suggestion
	return false
```

kkaneda: I see. Thanks!

PR #535, opened 2024-12-26
Cancel an inference request when an error occurs

Ladicle:


engine/internal/runtime/manager.go
```diff
@@ -290,16 +312,38 @@ func (m *Manager) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result,
...
+			log.Error(err, "Failed to list pods")
+			return ctrl.Result{}, err
+		}
+		for _, pod := range podList.Items {
+			if pod.Labels["controller-revision-hash"] != sts.Status.CurrentRevision {
+				continue
+			}
+			for _, cond := range pod.Status.Conditions {
+				if cond.Type == corev1.PodScheduled {
+					if cond.Status == corev1.ConditionFalse &&
```
kkaneda, line 330: Just wanted to make sure this works fine when a cluster is scaling up. The current request gets canceled, but a subsequent request will succeed once a pod is scheduled.

engine/internal/runtime/manager.go
```diff
@@ -290,16 +312,38 @@ func (m *Manager) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result,
...
+		); err != nil {
+			log.Error(err, "Failed to list pods")
+			return ctrl.Result{}, err
+		}
+		for _, pod := range podList.Items {
+			if pod.Labels["controller-revision-hash"] != sts.Status.CurrentRevision {
+				continue
+			}
+			for _, cond := range pod.Status.Conditions {
+				if cond.Type == corev1.PodScheduled {
```
kkaneda, line 329: No strong opinion, but we can combine these two `if`s into one.

kkaneda (approved):
LG!

engine/internal/runtime/manager.go
Ladicle: The first `if` is used for an early break.

engine/internal/runtime/manager.go
Ladicle: Yes, this only affects the waiting requests at that time, and subsequent requests do not change their behavior.

PR #536, opened 2024-12-27
feat: route request to another engine when the runtime fail to schedule

Ladicle:


engine/internal/runtime/manager.go
```diff
@@ -61,30 +61,32 @@ func newReadyRuntime(name string, address string) runtime {
...
+	ready     bool
+	errReason string
 	// address is empty when the runtime is not ready.
 	address string
 	// waitCh is used when the runtime is not ready.
 	waitCh chan struct{}
 }
 
-func (m *Manager) addRuntime(modelID string, sts appsv1.StatefulSet) (bool, error) {
+func (m *Manager) addRuntime(modelID string, sts appsv1.StatefulSet) (bool, bool, error) {
```
kkaneda, line 73: How about adding a comment to explain what the first `bool` and the second `bool` mean?

engine/internal/runtime/manager.go
```diff
@@ -228,11 +241,14 @@ func (m *Manager) PullModel(ctx context.Context, modelID string) error {
...
+	if m.errReason(modelID) != "" {
+		// The runtime is already in an error state.
+		return ErrRequestCanceled
+	}
 	log.Info("Waiting for runtime to be ready", "model", modelID)
 	select {
 	case <-r.waitCh:
-		ready, _ := m.isReady(modelID)
-		if !ready {
+		if m.errReason(modelID) != "" {
```
kkaneda, line 251: Slightly prefer making `errReason` return `string, bool`, but no strong opinion.

server/internal/router/route_map.go
```diff
@@ -94,18 +94,20 @@ func (r *routeMap) deleteEngine(engineID string) {
...
 
-	if len(r.engines) == 0 {
-		return "", fmt.Errorf("no engine available")
-	}
-
 	// modelsByID is a list of models keyed by engine IDs. This includes engines that don't have any models
 	modelsByID := make(map[string][]model)
 	for id := range r.engines {
-		modelsByID[id] = []model{}
+		if !ignores[id] {
```
kkaneda, line 104: I might not be following the logic correctly, but shouldn't we have a similar logic at line 115 as well?

engine/internal/runtime/manager.go
```diff
@@ -278,14 +294,23 @@ func (m *Manager) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result,
...
+		if ready, added, err := m.addRuntime(modelID, sts); err != nil {
 			log.Error(err, "Failed to add runtime")
 			return ctrl.Result{}, err
 		} else if added {
 			if err := m.autoscaler.Register(ctx, modelID, &sts); err != nil {
 				log.Error(err, "Failed to register autoscaler")
 				return ctrl.Result{}, err
 			}
+			if !ready {
+				if yes, err := hasUnschedulableChildren(ctx, m.k8sClient, sts); err != nil {
```
kkaneda, line 306: Question: Suppose that there are two pods for the statefulset, and one of the pods is unschedulable. In such a case, should we still cancel the request?

kkaneda (approved):
LG!

server/internal/router/route_map.go
Ladicle: Nice catch! L115 would be a problem when the ready model fails due to an unschedulable error.

engine/internal/runtime/manager.go
kkaneda: If so, shouldn't we change `hasUnschedulableChildren` so that it returns true only if all pods are unschedulable?

Ladicle: It's unnecessary, but changing it for better readability would be fine.

kkaneda: ah i see what you meant. No strong opinion!

PR #538, opened 2025-01-13
feat: bump vllm version to support Phi-4

kkaneda:


Ladicle (approved):

kkaneda:
vLLM issue: https://github.com/vllm-project/vllm/issues/11238

PR #563, opened 2025-02-12
feat(ingress): cover the path `/v1/inference/status` of `InferenceService`

takeshi-cloudnatix:


deployments/server/templates/ingress.yaml
```diff
@@ -42,3 +42,10 @@ spec:
             name: {{ include "inference-manager-server.fullname" . }}-http
             port:
               number: {{ .Values.httpPort }}
+      - path: /v1/inference
+        pathType: Prefix
+        backend:
+          service:
+            name: {{ include "inference-manager-server.fullname" . }}-http
```
kenji-cloudnatix, line 49: Actually I'm not sure if we use the same port for this. @guangrui-cloudnatix ?

guangrui-cloudnatix: I was thinking to use a different port, but not sure either

deployments/server/templates/ingress.yaml
```diff
@@ -42,3 +42,10 @@ spec:
             name: {{ include "inference-manager-server.fullname" . }}-http
             port:
               number: {{ .Values.httpPort }}
+      - path: /v1/inference
```
guangrui-cloudnatix, line 45: also per https://cloudnatix.atlassian.net/browse/CNATIX-5728, I thought the path would be `/v1/inference/status`?

deployments/server/templates/ingress.yaml
takeshi-cloudnatix: Ah, if there's more to do, please just ignore this PR for now. I just thought that we just need to add this.

PR #565, opened 2025-02-13
feat(server): setup status service

guangrui-cloudnatix:


kkaneda (approved):
LG!

cc @takeshi-cloudnatix

PR #567, opened 2025-02-14
feat(server,engine): implement inference status

guangrui-cloudnatix:
to be tested


server/internal/server/status.go
```diff
@@ -4,43 +4,82 @@ import (
...
 
 // RunWithListener runs the server with a given listener.
-func (s *ISS) RunWithListener(ctx context.Context, l net.Listener) error {
-	srv := grpc.NewServer()
+func (s *ISS) RunWithListener(ctx context.Context, authConfig config.AuthConfig, l net.Listener) error {
+	var opt grpc.ServerOption
+	if authConfig.Enable {
+		ai, err := auth.NewInterceptor(ctx, auth.Config{
+			RBACServerAddr: authConfig.RBACInternalServerAddr,
+			AccessResource: "api.inference.statuses",
```
guangrui-cloudnatix, line 70: not sure if this api name looks right? @kkaneda

guangrui-cloudnatix: actually not sure if need to create this resource?

kkaneda:
We can do in a separate PR, but as mentioned in the ticket we need to change https://github.com/llmariner/inference-manager/blob/main/server/internal/taskexchanger/task_receiver.go#L151 as well, I think.

server/internal/server/status.go
kkaneda: Ah I see. Yeah we can define a new resource, but we also then need to update rbac-server. For now, we can just have `"api.model"` and add a TODO comment to revisit later.

server/internal/server/status.go
```diff
@@ -59,7 +98,108 @@ func (s *ISS) Stop() {
 	s.srv.Stop()
 }
 
+// Refresh refreshes the inference status.
+func (s *ISS) Refresh(ctx context.Context, interval time.Duration) error {
```
kkaneda, line 102: A bit overkill to have caching/refreshing? I think it's ok to compute whenever a request comes. If we keep, can't we make the refresh period configurable?

server/internal/server/ws_server.go
```diff
@@ -156,6 +156,9 @@ func (ws *WS) processTasks(srv v1.InferenceWorkerService_ProcessTasksServer) err
 		switch msg := req.Message.(type) {
 		case *v1.ProcessTasksRequest_EngineStatus:
 			ws.logger.Info("Received engine status", "engineID", msg.EngineStatus.EngineId)
+			if msg.EngineStatus.EngineId == "" {
```
kkaneda, line 159: Question: When is `EngineId` empty?

engine/internal/runtime/manager.go
```diff
@@ -169,27 +208,31 @@ func (m *Manager) GetLLMAddress(modelID string) (string, error) {
...
 }
 
 // ListInProgressModels returns the list of models that are in progress.
-func (m *Manager) ListInProgressModels() []string {
+func (m *Manager) ListInProgressModels() []*ModelRuntimeInfo {
 	return m.listModels(false)
 }
 
-func (m *Manager) listModels(ready bool) []string {
+func (m *Manager) listModels(ready bool) []*ModelRuntimeInfo {
```
kkaneda, line 221: Just wanted to make sure `ModelRuntimeInfo` is immutable since we return them to a caller without copying.

engine/internal/runtime/manager.go
```diff
@@ -48,44 +48,83 @@ type Manager struct {
...
 		if err != nil {
 			return false, false, err
 		}
-		m.runtimes[modelID] = newReadyRuntime(sts.Name, c.GetAddress(sts.Name))
+		rt = newReadyRuntime(sts.Name, c.GetAddress(sts.Name), sts.Status.ReadyReplicas)
 	} else {
-		m.runtimes[modelID] = newPendingRuntime(sts.Name)
+		rt = newPendingRuntime(sts.Name)
 	}
+	rt.gpu = getGPU(sts)
```
kkaneda, line 126: I believe we need to update the gpu count from `Reconcile` when the statefulset is scaled up/down. Can you add a TODO comment if that's correct?

kkaneda (commented):
nice!

Ladicle (approved):

engine/internal/runtime/manager_test.go
```diff
@@ -127,30 +129,36 @@ func TestGetLLMAddress(t *testing.T) {
...
-	assert.Contains(t, models, "model-0")
-	assert.Contains(t, models, "model-2")
+	for _, m := range models {
+		contains := m.ID == "model-0" || m.ID == "model-2"
+		assert.True(t, contains)
+	}
 }
 
-func TestListProgressModelIDs(t *testing.T) {
+func TestListProgressModels(t *testing.T) {
```
takeshi-cloudnatix, line 148: how about `TestListInProgressModels`

takeshi-cloudnatix (approved):
Thank you!

engine/internal/runtime/manager.go
guangrui-cloudnatix: `rt.gpu` is gpu number specified in container spec, ref. line 79. 

the number of replicas is synced periodically from engine to server. server computes and reports the total gpu allocated by 'rt.gpu * replicas`

kkaneda: Ah I see. But still `replicas` needs to be updated when the statefulset is autoscaled?

kkaneda: Let me check my understanding of the code.

- `sendEngineStatus` reports  `m.GPU` .
- `m.GPU` is calculated by `r.gpu * r.replicas`.
- `r.replicas` is set when `newReadyRuntime` is called, and `r.replicas` is not updated after that.

The scenario I was thinking of is following:

1. `inference-engine-manager` creates a new statefulset (replica size = 2).
2. The statefulset becomes ready when one of the pods become ready. The other pod is still not ready.
3. `newReadyRuntime` is called.
4. The other pod of the statefulset becomes ready. `r.replicas` should be updated to 2, but it is not.

Another scenario is:

1. `inference-engine-manager` creates a new statefulset (replica size = 1).
2. The statefulset becomes ready. The pod becomes ready.
3. `newReadyRuntime` is called.
4. The statefulset is scaled up, and a new pod is created. `r.replicas` should be updated to 2, but it is not.

Even the above is true, we just want to have a TODO comment now. We would first need to focus on the end-to-end flow. (I'm now thinking that the initial guideline I put in the Jira ticket might not work well when there is more than one engine in the cluster.)

guangrui-cloudnatix: The two scenarios above are for newly created statefulset
At line 382 of this file, `ReadyReplicas` is updated periodically for existing statefulset.

engine/internal/runtime/manager.go
```diff
@@ -336,7 +379,7 @@ func (m *Manager) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result,
 			log.Error(err, "Still unable to reach the runtime endpoint", "retry-after", retryAfter)
 			return ctrl.Result{RequeueAfter: retryAfter}, err
 		}
-		m.markRuntimeReady(sts.Name, modelID, addr)
+		m.markRuntimeReady(sts.Name, modelID, addr, sts.Status.ReadyReplicas)
```
guangrui-cloudnatix, line 382: @kkaneda  For existing statefulsets, theirs active replicas should be updated here periodically

engine/internal/runtime/manager.go
kkaneda: Ah I don't think so. 

[Line 334](https://github.com/llmariner/inference-manager/blob/ad0352287494cbdde197f4c6bbca0301372dd118/engine/internal/runtime/manager.go#L334) sets `ready` to `true` if a statefulset is already ready. We hit line 382 only when `ready` is `false` ([Line 360](https://github.com/llmariner/inference-manager/blob/ad0352287494cbdde197f4c6bbca0301372dd118/engine/internal/runtime/manager.go#L360)).

guangrui-cloudnatix:
sample output:
```
grpcurl -plaintext localhost:8091 llmariner.inference.server.v1.InferenceService.ListInferenceStatus
{
  "clusterStatuses": [
    {
      "id": "default",
      "engineStatuses": [
        {
          "engineId": "engine_VuY4S-HVJ8Fy3eKpQNoHHMG2",
          "models": [
            {
              "id": "google-gemma-2b-it-q4_0",
              "isReady": true,
              "inProgressTaskCount": 1,
              "gpuAllocated": 1
            }
          ],
          "clusterId": "default"
        }
      ]
    }
  ],
  "taskStatus": {
    "inProgressTaskCounts": {
      "google-gemma-2b-it-q4_0": 1
    }
  }
}
```

deployments/server/values.yaml
```diff
@@ -385,6 +385,10 @@ securityContext:
   runAsNonRoot: true
   runAsUser: 1000
 
+# How often (in seconds) to refresh inference status. Default to 60 seconds.
+# +docs:type=number
+statusRefreshInterval: 60
```
kkaneda, line 390: This should be `60s` or `1m` since the type is `time.Duration`.

engine/internal/runtime/manager.go
kkaneda: does the above comment make sense?

kkaneda (approved):
LGTM!

engine/internal/runtime/manager.go
guangrui-cloudnatix: Yes, and added line 378 to cover the case

PR #569, opened 2025-02-18
feat(api): Support max_completion_tokens in CreateChatCompletionRequest

alpe:
Resolves https://github.com/llmariner/llmariner/issues/622

This implements option 1)

* Add `max_completion_tokens` to the `CreateChatCompletionRequest`
* proto generation and  buf format
* Introduce test fixture and minor refactoring in test


buf.lock
```diff
@@ -4,4 +4,5 @@ deps:
   - remote: buf.build
     owner: googleapis
     repository: googleapis
-    commit: 83c0f6c19b2f4ea0b0fd84a80e753659
+    commit: 546238c53f7340c6a2a6099fb863bc1b
+    digest: shake256:8d75c12f391e392b24c076d05117b47aeddb090add99c70247a8f4389b906a65f61a933c68e54ed8b73a050b967b6b712ba194348b67c3ab3ee26cc2cb25852c
```
alpe, line 8: This dependency failed without a digest on my box. I was using `v1.50.0`

server/internal/server/completions_test.go
```diff
@@ -46,231 +46,141 @@ func TestCreateChatCompletion(t *testing.T) {
...
-								Type: "text",
-								Text: "test",
-							},
-						},
-					},
-				},
-			},
-			code: http.StatusOK,
+			name:    "success",
+			req:     CreateChatCompletionRequestFixture(WithModel(modelID)),
```
alpe, line 62: I have introduced a fixture for the common request defaults. The old test setup was fine but this makes it a bit more readable IMHO as the changes are more visible.

server/internal/server/completions_test.go
```diff
@@ -46,231 +46,141 @@ func TestCreateChatCompletion(t *testing.T) {
...
+			expCode: http.StatusBadRequest,
+		},
+		{
+			name: "MaxCompletionTokens set no MaxTokens",
+			req: CreateChatCompletionRequestFixture(WithModel(modelID), func(c *v1.CreateChatCompletionRequest) {
+				c.MaxCompletionTokens = 200
+			}),
+			expCode: http.StatusOK,
+			assert: func(t *testing.T, req *v1.CreateChatCompletionRequest) {
+				assert.Equal(t, int32(200), req.MaxCompletionTokens)
```
alpe, line 177: This way, I have confidence that the request was modified as expected.

server/internal/server/completions_test.go
```diff
@@ -46,231 +46,141 @@ func TestCreateChatCompletion(t *testing.T) {
...
+			expCode: http.StatusBadRequest,
+		},
+		{
+			name: "MaxCompletionTokens not set - use MaxTokens as default",
+			req: CreateChatCompletionRequestFixture(WithModel(modelID), func(c *v1.CreateChatCompletionRequest) {
+				c.MaxTokens = 100
+			}),
+			expCode: http.StatusOK,
+			assert: func(t *testing.T, req *v1.CreateChatCompletionRequest) {
+				assert.Equal(t, int32(100), req.MaxCompletionTokens)
```
alpe, line 148: With the modified request captured, I have confidence that my code was executed

server/internal/server/fixtures_test.go
```diff
@@ -0,0 +1,52 @@
+package server
```
alpe, line 1: A better file name would be `test_fixtures.go` . How do you feel about making the functions public?
They would be still within the `internal` package and not visible when used as a library.

common.mk
```diff
@@ -58,7 +58,7 @@ buf-generate-all: buf-generate-base buf-generate-ts
 
 .PHONY: buf-mod-update
 buf-mod-update:
-	buf mod update
+	buf dep update
```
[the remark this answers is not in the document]
kkaneda: nice catch! How about updating the make target as well (to `buf-dep-update`)?

server/internal/server/completions.go
```diff
@@ -127,7 +127,13 @@ func (s *S) CreateChatCompletion(
 			IncludeUsage: true,
 		}
 	}
-
+	// max_tokens is deprecated and replaced by max_completion_tokens, now
+	if createReq.MaxCompletionTokens == 0 {
```
kkaneda, line 131: How about also having `createReq.MaxTokens = createReq.MaxCompletionTokens` when `createReq.MaxTokens == 0` and adding a (TODO) comment mentioning that this is required as Ollama doesn't support `MaxCompletionTokens` as of Feb 2025?

server/internal/server/fixtures_test.go
kkaneda: Looks good! If we really want to make these work only in test we can change the functions to take `*testing.T` as a argument, but that would be probably overkill.

server/internal/server/completions.go
alpe: good idea!

PR #590, opened 2025-02-22
feat: add more per-cluster status

kkaneda:


server/internal/server/status_test.go
```diff
@@ -87,6 +87,8 @@ func TestGetInferenceStatus(t *testing.T) {
 								},
 							},
 						},
+						ModelCount:          3,
+						InProgressTaskCount: 24,
```
takeshi-cloudnatix, line 91: just to make sure, is it fine to take only 10 on engine2 for model1 or we should fix the fake input above?

takeshi-cloudnatix (commented):
LGTM!

server/internal/server/status_test.go
kkaneda: nice catch! let me fix.

PR #593, opened 2025-02-24
fix: fix the calculation of per-cluster in-flight request

kkaneda:
Sum up in-flight requests from all engines.


Ladicle (approved):

takeshi-cloudnatix (commented):
LGTM!

PR #594, opened 2025-02-24
fix: use normal distroless image as a base image for the engine

Ladicle:
The nonroot image does not have permission to write ebs volume. We can set the securityContext and change group, but it leads to permission mismatch to vllm image.


kenji-cloudnatix (approved):
I see. LG!

PR #595, opened 2025-02-24
Update the protobuf generated files with ones generated by using the …

takeshi-cloudnatix:
…latest `buf`

`common.mk` in this repo runs `buf dep update`. The `dep` command was introduced from https://github.com/bufbuild/buf/releases/tag/v1.32.0.

Though we have https://github.com/llmariner/llmariner/blob/main/DEVELOPMENT.md where `v1.5.0` is suggested, this means that the files were generated using `v1.32.0` or later.

I ran `make generate` using Buf `v1.50.0` using this opportunity after running:

```
go install github.com/bufbuild/buf/cmd/buf@v1.5.0 && \
  go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.34.1 && \
  go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.1.0 && \
  go install github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway@v2.6.0 && \
  go install github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2@v2.6.0 && \
  go install github.com/grpc-ecosystem/protoc-gen-grpc-gateway-ts@v1.1.1
```

I tested the following releases:
- `v1.31.0`: doesn't build
- `v1.32.0`: builds
- `v1.33.0`, `v1.35.0`, `v1.40.0`, `v1.45.0`, `v1.48.0`, `v1.49.0`, `v1.50.0`: The same as `v1.32.0` except that `EnumValueOptions` will have the `featureSupport` field

PR #596, opened 2025-02-24
Generate TypeScript files from the protobuf with the `use_proto_names…

takeshi-cloudnatix:
…` option turned on

By default, `protoc-gen-grpc-gateway-ts` generates files with field names named by using the lower camel case.

Some of our servers are using gRPC-Gateway v2 with `UseProtoNames` set to `true`.
https://github.com/search?q=org%3Allmariner%20UseProtoNames&type=code

See also https://github.com/grpc-ecosystem/grpc-gateway/blob/2ee9ba58097ac79e34f3e63beef9358d4f24d2fd/docs/docs/development/grpc-gateway_v2_migration_guide.md#we-now-use-the-camelcase-json-names-by-default

To make the generated TypeScript files work with it, we need to use the `use_proto_names` option.
https://github.com/grpc-ecosystem/protoc-gen-grpc-gateway-ts?tab=readme-ov-file#use_proto_names

Currently, we're manually copying what's needed from generated files to `llmariner/frontend` repo (under the `src/next/types/` directory) after manually editing them to fix the naming convention. This manual editing step can be eliminated by just using the `use_proto_names` option.

From the beginning, maybe we can revisit the use of `UseProtoNames` option on the server side.

Note that I used Buf `v1.50.0` to run `make generate` which is different from https://github.com/llmariner/llmariner/blob/main/DEVELOPMENT.md#protocol-buffers.

See https://github.com/llmariner/inference-manager/pull/595#issue-2875189503 for the reason.

### DIFF
--- a/buf.gen.ts.yaml
+++ b/buf.gen.ts.yaml
@@ -2,3 +2,5 @@ version: v1
 plugins:
 - name: grpc-gateway-ts
   out: ./ts
+  opt:
+  - use_proto_names=true
--- a/dist/api/v1/inference_server.pb.d.ts
+++ b/dist/api/v1/inference_server.pb.d.ts
@@ -21,16 +21,16 @@ export type CreateChatCompletionRequestMessageContentAudioUrl = {
 export type CreateChatCompletionRequestMessageContent = {
     type?: string;
     text?: string;
-    imageUrl?: CreateChatCompletionRequestMessageContentImageUrl;
-    inputAudio?: CreateChatCompletionRequestMessageContentInputAudio;
-    audioUrl?: CreateChatCompletionRequestMessageContentAudioUrl;
+    image_url?: CreateChatCompletionRequestMessageContentImageUrl;
+    input_audio?: CreateChatCompletionRequestMessageContentInputAudio;
+    audio_url?: CreateChatCompletionRequestMessageContentAudioUrl;
 };
 export type CreateChatCompletionRequestMessage = {
     content?: CreateChatCompletionRequestMessageContent[];
     role?: string;
     name?: string;
-    toolCalls?: CreateChatCompletionRequestMessageToolCall[];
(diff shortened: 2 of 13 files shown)


kkaneda:
Ah sorry for the confusion on `buf` versions! We have updated the command with https://github.com/llmariner/inference-manager/pull/569/files .

> I've been doing this, but this might be not true for you. Let me know if you're not doing what's described here.

I see. I never updated the files there. Maybe Guangrui or Saniya did, but not sure.

> Let me know the reason behind UseProtoNames if any.

This is mainly to follow the OpenAI API spec (https://github.com/llmariner/inference-manager/pull/569/files). We don't need to follow some of our own endpoints though.

kkaneda (approved):

takeshi-cloudnatix:
> Ah sorry for the confusion on `buf` versions! We have updated the command with https://github.com/llmariner/inference-manager/pull/569/files .

Ah, I see! Switching to `dep` makes sense. Let me update `DEVELOPMENT.md`.

> > I've been doing this, but this might be not true for you. Let me know if you're not doing what's described here.
> 
> I see. I never updated the files there. Maybe Guangrui or Saniya did, but not sure.

Got it!

> > Let me know the reason behind UseProtoNames if any.
> 
> This is mainly to follow the OpenAI API spec (https://github.com/llmariner/inference-manager/pull/569/files). We don't need to follow some of our own endpoints though.

Ah, understood.

takeshi-cloudnatix:
> > Ah sorry for the confusion on `buf` versions! We have updated the command with https://github.com/llmariner/inference-manager/pull/569/files .
> 
> Ah, I see! Switching to `dep` makes sense. Let me update `DEVELOPMENT.md`.

addressing by https://github.com/llmariner/llmariner/pull/677/files

PR #602, opened 2025-03-11
feat(api): generate `index.js` and `index.d.ts` in the `dist` directory

takeshi-cloudnatix:
See https://github.com/llmariner/cluster-manager/pull/70#discussion_r1988404679 for the details.


ts/index.ts
takeshi-cloudnatix, line 1: excluded `inference_server_internal`

PR #609, opened 2025-03-18
feat(engine): add puller daemon mode to the runtime client

Ladicle:
This PR only adds setting values and supports the deployment of the puller in the daemon mode. It does not yet interact with the existing runtime manager.


engine/cmd/pull.go
```diff
@@ -127,26 +119,23 @@ func runServer(ctx context.Context, c *config.Config, socketPath string) error {
...
 		defer cancel()
-		log.Printf("Starting HTTP server at %q\n", socketPath)
-		if err := srv.Serve(l); err != http.ErrServerClosed {
+		log.Printf("Starting HTTP server at %q\n", srv.Addr)
+		if err := srv.ListenAndServe(); err != http.ErrServerClosed {
 			log.Printf("serve: %v\n", err)
 		}
 	}()
 
+	if modelID != "" {
```
kkaneda, line 135: Just wanted to check my understanding. When the deamon mode is enabled, specifying a model ID is optional. When a model ID is specified, the model is pulled at the deamon startup time. 

Maybe nice to have some comment somewhere.

engine/internal/runtime/client.go
```diff
@@ -124,7 +128,12 @@ func (c *commonClient) deployRuntime(
 	update bool,
 ) (*appsv1.StatefulSet, error) {
 	mci := c.mconfig.ModelConfigItem(params.modelID)
-	name := resourceName(mci.RuntimeName, params.modelID)
+	var name string
+	if params.pullerDaemonMode {
+		name = resourceName(mci.RuntimeName, "dynamic")
```
kkaneda, line 133: Slightly prefer defining a const for `dynamic` and use here and `ollama_client.go`.

engine/internal/config/config.go
```diff
@@ -33,6 +33,13 @@ type OllamaConfig struct {
 	Debug bool `yaml:"debug"`
 
 	RunnersDir string `yaml:"runnersDir"`
+
+	// DynamicModelLoading is true if the model is loaded dynamically.
+	// If this is set to true, the puller is run in the daemon mode.
+	DynamicModelLoading bool `yaml:"dynamicModelLoading"`
+	// PullerPort is the port for the puller. This is only used when
+	// DynamicModelLoading is true.
+	PullerPort int `yaml:"pullerPort"`
```
kkaneda, line 42: How about checking if `PullerPort` is > 0 when `DynamicModelLoading` is true?

kkaneda (approved):
LG!

engine/cmd/pull.go
Ladicle: Yes, that's correct. I'll add a comment here.

PR #625, opened 2025-04-01
fix: run 'ollama create' for newly pulled models in the dynamic loading mode

kkaneda:


Ladicle (approved):
Just a question, are these changes for GGUG type models?

kenji-cloudnatix:
Ah good point. Yes, I think so. Let me investigate some more.

kkaneda:
Thanks! You're right. This is only needed for GGUF (and probably HuggingFace format as well). Let me add a comment to clarify.

PR #667, opened 2025-04-17
feat(dist): re-build `dist` files

takeshi-cloudnatix:
using Buf version:

```
buf --version
1.52.1
```

then running

```
make generate
```

and running:

```
yarn build
```


takeshi-cloudnatix:
ah, never mind.

takeshi-cloudnatix:
not sure what was wrong, but some definitions are missing. This PR adds them.

takeshi-cloudnatix:
Hmm, I thought `yarn build` was needed, but it seems `make generate` is enough for generating new files.

`make generate` should be with `rm -rf ./dist`...

takeshi-cloudnatix:
> Hmm, I thought `yarn build` was needed, but it seems `make generate` is enough for generating new files.
> 
> `make generate` should be with `rm -rf ./dist`...

fixing by https://github.com/llmariner/inference-manager/pull/684

PR #684, opened 2025-04-18
fix(dist): fix `index.js` and `index.d.ts` to re-export the up-to-dat…

takeshi-cloudnatix:
…e definitions

https://github.com/llmariner/inference-manager/pull/667 was incomplete. I had to update the non-generated file `index.ts` to import the right files from the `api/v1` directory.


dist/api/v1/inference_server_status.pb.d.ts
takeshi-cloudnatix, line 1: The source of these file is now gone.

common.mk
```diff
@@ -73,7 +73,8 @@ endif
 .PHONY: typescript-compile
 typescript-compile: buf-generate-ts
 ifneq (,$(wildcard ./buf.gen.ts.yaml))
-	tsc --skipLibCheck
```
takeshi-cloudnatix, line 76: I think we don't need the `--skipLibCheck` option here. removing.

common.mk
```diff
@@ -73,7 +73,8 @@ endif
...
+	rm -fr ./dist
```
takeshi-cloudnatix, line 76: porting this from `package.json`

ts/api/v1/inference_server_status.pb.ts
takeshi-cloudnatix, line 1: The source proto file for these files is gone.

We can add the `--clean` flag (https://buf.build/docs/reference/cli/buf/generate/#flags) to the `buf generate` command to delete old files in the `ts` directory automatically, before generation. But it also deletes the manually-crafted file `ts/index.ts` which we need to keep.

We can add some complex command using e.g. `find` to achieve that, but for now, let me keep it as a manual task.

package.json
```diff
@@ -9,21 +9,12 @@
     "ts",
     "/dist"
   ],
-  "devDependencies": {
-    "typescript": "^4.5.4"
-  },
-  "scripts": {
-    "build": "rm -rf ./dist && tsc"
```
takeshi-cloudnatix, line 16: If `yarn` is run, the `tsc` binary from the `typescript` dep is available at `node_modules/.bin`. When this script is run, `node_modules/.bin` is added to the `PATH` env. So, the `tsc` binary of the `typescript` dep will be used.

This is good for controlling what binary to use, but for code generation from the protobuf files, we don't have any automation.

See https://github.com/llmariner/llmariner/blob/main/DEVELOPMENT.md#protocol-buffers.

For now, let me leave `tsc` the same as the protobuf toolchain (each developer should prepare it).

package.json
```diff
@@ -9,21 +9,12 @@
...
-  },
   "repository": {
     "type": "git",
     "url": "git+https://github.com/llmariner/inference-manager.git"
   },
   "author": "llmariner",
   "license": "Apache-2.0",
-  "bugs": {
```
takeshi-cloudnatix, line 24: Let me remove this using this opportunity. "Issues" is disabled for https://github.com/llmariner/inference-manager.

https://github.com/llmariner/inference-manager/issues is working though it's disabled (not listed on the bar). Users who noticed this may go there and file an issue there while it's intended to be unavailable. It's not good.

We can choose to put `https://github.com/llmariner/llmariner/issues` here instead. But anyway, we're not publishing these to the NPM/Yarn repository. I'd like to choose to not putting metadata that are not really necessary at this point.

kkaneda (approved):
nice!

PR #779, opened 2025-05-09
fix: set NeedLeaderElection to false in LoRAReconciler

kkaneda:


engine/internal/runtime/lora_manager.go
```diff
@@ -95,6 +95,11 @@ func (r *LoRAReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(r)
 }
 
+// NeedLeaderElection implements LeaderElectionRunnable.
+func (p *LoRAReconciler) NeedLeaderElection() bool {
```
kkaneda, line 99:
```suggestion
func (r *LoRAReconciler) NeedLeaderElection() bool {
```

PR #879, opened 2025-07-09
Ingress host

greg80303:
Add support for host name on Ingress rules


deployments/server/values.yaml
```diff
@@ -17,6 +17,10 @@ global:
     # +docs:property
     # annotations: {}
 
+    If provided, this value will be added to each rule of every Ingress
```
greg80303, line 20:
```suggestion
    # If provided, this value will be added to each rule of every Ingress
```

kkaneda (approved):

PR #961, opened 2025-08-14
chore: move the helper functions in runtime/client.go to the bottom

kkaneda:


engine/internal/runtime/client.go
```diff
@@ -512,6 +432,8 @@ func (c *commonClient) deployRuntime(
 		annos[modelAnnotationKey] = params.modelID
 	}
 
+	// Set WithUpdateStrategy(appsv1apply.StatefulSetUpdateStrategy().
```
kkaneda, line 435:
```suggestion
```

engine/internal/runtime/client.go
```diff
@@ -512,6 +432,7 @@ func (c *commonClient) deployRuntime(
 		annos[modelAnnotationKey] = params.modelID
 	}
 
+	// - Set minAvailable to zero
```
kkaneda, line 435:
```suggestion
```

PR #1061, opened 2025-09-18
feat(engine): shutdown vllm engine gracefully

guangrui-cloudnatix:


engine/internal/runtime/client.go
```diff
@@ -373,6 +373,25 @@ func (c *commonClient) deployRuntime(
 	if p := params.runtimePort; p != 0 {
 		cport = p
 	}
+
+	cmd := fmt.Sprintf(`
```
kkaneda, line 377: Can't we do this only for vLLM? We can add a new parameter to `deployRuntimeParams` and set the field only in the `vLLMClient`.

engine/internal/runtime/client.go
```diff
@@ -373,6 +373,25 @@ func (c *commonClient) deployRuntime(
...
+sleep 15
+while true; do
+	metrics=$(curl -s http://localhost:%d/metrics/)
+	running=$(echo "$metrics" | grep '^vllm:num_requests_running' | awk '{print $2}')
+	waiting=$(echo "$metrics" | grep '^vllm:num_requests_waiting' | awk '{print $2}')
+	if [[ "$running" == "0.0" && "$waiting" == "0.0" ]]; then
+	break
+	fi
+	sleep 1
```
kkaneda, line 386: I was wondering if we should print some log, but not too sure if we can see the log somewhere.

kkaneda (approved):
Cool! LG!

engine/internal/runtime/client.go
guangrui-cloudnatix: I had an echo line but I didn't see it printed out from `kubectl logs inference-engine ....`, so I removed the echo line

kkaneda: Got it! This is from Claude. Probably hallucination:

image

PR #1083, opened 2026-04-16
feat(engine): make the HuggingFace downloader download a file named `processor_config.json` if any

takeshi-cloudnatix:
needed to support Gemma 4


engine/internal/modeldownloader/huggingface/download.go
```diff
@@ -79,6 +79,7 @@ func DownloadModelFiles(ctx context.Context, s3Client s3Client, adapterType mv1.
 		{name: "tokenizer.json", isOptional: true},
 		{name: "tokenizer_config.json", isOptional: false},
 		{name: "preprocessor_config.json", isOptional: true},
+		{name: "processor_config.json", isOptional: true},
```
takeshi-cloudnatix, line 82: `google/gemma-4-31B-it` is using `processor_config.json`. We need to make the engine download it.

https://huggingface.co/google/gemma-4-31B-it/tree/main

